PR #51, opened 2025-10-28
Adds first draft of QueryLanguage

FriedJannik:
- Currently only Comparison Conditions are possible
- Only submodel idShort is supported for comparison

Co-authored-by: Aaron Zielstorff <aaron.zielstorff@iese.fraunhofer.de>

### DIFF
--- a/cmd/submodelrepositoryservice/aas_query.json
+++ b/cmd/submodelrepositoryservice/aas_query.json
@@ -2,8 +2,8 @@
     "Query": {
         "$condition": {
             "$eq": [
-                { "$field": "$sm#id" },
-                { "$strVal": "1" } 
+                { "$field": "$sm#semanticId.keys[1].type" },
+                { "$strVal": "Property" } 
             ]
         }
     }
--- a/cmd/submodelrepositoryservice/main.go
+++ b/cmd/submodelrepositoryservice/main.go
@@ -69,7 +69,7 @@ func runServer(ctx context.Context, configPath string, databaseSchema string) er
 		}
 	}()
 
-	//submodelrepository.TestNewSubmodelHandler(smDatabase)
+	// submodelrepository.TestNewSubmodelHandler(smDatabase)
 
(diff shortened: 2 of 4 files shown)


internal/common/model/querylanguage/model_comparison.go
Martin187187: The access control model contains logic like this aswell because the acm contains aas query logic aswell.
The acm has Logical Expressions: https://github.com/Martin187187/basyx-go-components/blob/discoveryservice-security/internal/common/security/model/access_rule_model.go
and eval.go has the validation logic: https://github.com/Martin187187/basyx-go-components/blob/discoveryservice-security/internal/common/security/eval.go

Can we combine this logic or are there slight differences

internal/common/model/querylanguage/model_query.go
```diff
@@ -0,0 +1,246 @@
...
+			q.Condition = &logicalExpr
+		} else {
+			return fmt.Errorf("unknown condition type")
+		}
+	}
+
+	return nil
+}
+
+func HandleFieldToValueComparison(leftOperand, rightOperand Operand, operation string) (exp.Expression, error) {
```
Martin187187: Next 3 functions are extremely similar. Can this be combined?

internal/common/model/querylanguage/model_logical_expression.go
```diff
@@ -0,0 +1,268 @@
...
+	operands := operationObj.GetOperands()
+	if len(operands) != 2 {
+		return nil, fmt.Errorf("comparison operation requires exactly 2 operands, got %d", len(operands))
+	}
+
+	leftOperand := operands[0]
+	rightOperand := operands[1]
+
+	// Handle different operand type combinations
+	if leftOperand.GetOperandType() == "$field" && rightOperand.GetOperandType() != "$field" {
```
Martin187187: all functions here are extremely similar here and have same parameters aswell. Maybe simplify?

PR #55, opened 2025-10-30
Refactors SubmodelRepository Handler/Database

FriedJannik:
# Refactor of SubmodelRepository Handler/Database
## Description of Changes

This PR removes unused and prototype-level code. It got replaced by the new SubmodelHandler that is already part of the codebase.
Note: SubmodelElements are still work-in-progress and are excluded from all Submodel specific endpoints. Submodel Element Endpoints should still work as expected.

I also Removed the Read Method from the SubmodelElement Database Handlers - why? There is no need to have a specific type-related Read operation as type handling must be done in the SQL query for performance reasons - this is to follow our "Single-Query" approach, also they were unused anyway.

## Related Issue

--

--

## AAS Files Used for Testing

--

## Additional Information

--


internal/submodelrepository/persistence/SubmodelElements/SMECrudInterface.go
FriedJannik: This file is not new - it was moved and changed (Removed Read method from interface)

aaronzi (approved):

PR #69, opened 2025-11-05
Lint setup

Martin187187:


FriedJannik:
# On Behalf of the Line Length:
- Go has no official max line length (unlike e.g. Python’s PEP 8).
- Automatically breaking lines can change readability or semantics:
   - function calls
   - composite literals
   - chained calls
   - comments vs code

- The Go team considers line wrapping a human judgment, not a mechanical one.

So: **No Line Length Linter**

PR #87, opened 2025-11-17
Aasregistry security

Martin187187:
## Goal
- Enforce Attribute-Based Access Control (ABAC) consistently across the AAS Registry service so every request routes through the same descriptor filters and logical-expression evaluation.
- iplements: https://github.com/orgs/eclipse-basyx/projects/6/views/1?pane=issue&itemId=139794563

## Key Additions
- `cmd/aasregistryservice` and the security package now wrap endpoint handlers in the ABAC engine so descriptor queries only return allowed assets or descriptors. This makes every HTTP endpoint subject to a uniform permissions check rather than relying on ad-hoc guards.
- The ABAC model gained richer grammar support (`internal/common/model/grammar/*`), including object/attribute definitions, rights enums, logical expressions, and ACL helpers so descriptors can be filtered directly in the query language before evaluating ABAC rules.
- Descriptor-aware evaluation logic now lives inside `internal/common/security/abac_engine.go` and related unit tests. The engine evaluates descriptor data directly (not just rule metadata) while handling AND/OR reductions, unknown fields, and missing filters, which stabilizes permission checks for complex descriptor scenarios.
- Query parsing for descriptors was extended (`query.go`, `query_test.go`) with improved join handling and support for multiple descriptor objects, enabling the security layer to reason about specific descriptor substructures.

## Testing & Validation
- Added focused test cases for the ABAC engine and grammar/model components (`internal/common/security/unit_tests/abac_engine`, `internal/common/security/unit_tests/model/testdata`), covering success/failure cases such as anonymous write attempts, wrong data types, and logical-expression evaluation.

### DIFF
--- a/cmd/aasregistryservice/config.yaml
+++ b/cmd/aasregistryservice/config.yaml
@@ -18,7 +18,7 @@ oidc:
   audience: "discovery-service"
 
 abac:
-  enabled: true
+  enabled: false
   enableDebugErrorResponses: true
   clientRolesAudience: "discovery-service"
   modelPath: "config/access_rules/access-rules.json"


export/master-realm.json
Martin187187: delete

docu/security/DISCOVERY_AUTHORIZATION_deprecated.md
aaronzi: Why does the title of the file say deprecated?

internal/aasregistry/api/api_asset_administration_shell_registry_api_service.go
```diff
@@ -8,8 +32,8 @@
  * API version: V3.1.1_SSP-001
  * Contact: info@idtwin.org
  */
-
-package ass_registry_api
+// Author: Martin Stemmer ( Fraunhofer IESE )
+package assregistryapi
```
aaronzi: please rename to aasregistryapi

internal/common/descriptors/SubModelDescriptorHandler.go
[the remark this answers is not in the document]
Martin187187: Not sure if that is needed. A /shell-descriptors would fetch ALL submodel descriptors without limit. So doing it in memory is nothing different. here

Martin187187:
Massive logical fixes:

- Corrected position-based filtering: previous logic failed when multiple rows with different positions existed, because all conditions ran on the same row. Now each condition uses its own EXISTS block, so different rows can satisfy different constraints.

- Improved safe casting: values that cannot be cast to the target SQL type are now ignored instead of causing cast errors. This required additional internal regex guards and CASE logic to prevent invalid casts.

PR #110, opened 2026-01-05
Registryofregistries

chris-koo:
## Description of Changes

- The /registry-descriptors endpoint now allows filtering for the endpoint interface.  
- Integration tests for the registry of registries have been added.

### DIFF
--- a/examples/BaSyxRegistryOfRegistries/startup.sh
+++ b/examples/BaSyxRegistryOfRegistries/startup.sh
@@ -102,7 +102,7 @@ curl --location --request POST 'http://localhost:5080/registry-descriptors' \
     "endpoints": [
         {
             "protocolInformation": {
-                "href": "https://demo.digital-twin.host/aas-discovery/lookup/shells",
+                "href": "https://demo.digital-twin.host/aas-discovery",
                 "endpointProtocol": "HTTPS"
             },
             "interface": "AAS-DISCOVERY-3.0"
--- a/internal/registryofregistriesservice/integration_tests/expected/registry_descriptor_cursor_limit.json
+++ b/internal/registryofregistriesservice/integration_tests/expected/registry_descriptor_cursor_limit.json
@@ -33,7 +33,7 @@
                 {
                     "interface": "AAS-DISCOVERY-3.0",
                     "protocolInformation": {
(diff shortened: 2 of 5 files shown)


internal/common/descriptors/RegistryDescriptorHandler.go
```diff
@@ -355,6 +357,17 @@ func ListRegistryDescriptors(
 		ds = ds.Where(reg.Col(colCompany).Eq(company))
 	}
 
+	if endpointInterface != "" {
+		ds = ds.
+			Join(
```
Martin187187: Left join

internal/registryofregistriesservice/integration_tests/integration_test.go
```diff
@@ -0,0 +1,298 @@
...
+				}
+
+				assert.JSONEq(t, expectedStr, responseStr, "Response does not match expected")
+				t.Logf("Expected: %s", expectedBytes)
+			}
+
+			t.Logf("Response: %s", response)
+		})
+	}
+}
```
Martin187187: can you add this as discussed:
```
t.Run("Check_DB_Empty", func(t *testing.T) {
	       db, err := sql.Open("postgres", "host=127.0.0.1 port=5432 user=admin password=admin123 dbname=basyxTestDB sslmode=disable")
	       require.NoError(t, err)
	       defer func() { _ = db.Close() }()
	       require.NoError(t, db.Ping())

	       rows, err := db.Query("SELECT tablename FROM pg_tables WHERE schemaname = 'public'")
	       require.NoError(t, err)
	       defer func() { _ = rows.Close() }()

	       nonEmpty := []string{}
	       for rows.Next() {
	           var table string
	           require.NoError(t, rows.Scan(&table))

	           var cnt int
	           q := fmt.Sprintf("SELECT COUNT(*) FROM \"%s\"", table)
	           err = db.QueryRow(q).Scan(&cnt)
	           require.NoError(t, err)
	           if cnt != 0 {
	               nonEmpty = append(nonEmpty, fmt.Sprintf("%s:%d", table, cnt))
	           }
	       }
	       require.NoError(t, rows.Err())

	       assert.Empty(t, nonEmpty, "Expected all tables empty, but found rows in: %v", nonEmpty)
	   })
```

PR #135, opened 2026-01-20
Registry Query Endpoint

Martin187187:
implements #146 

all you have to add in a query endpoint is:
image


FriedJannik (approved):
LGTM. about the QueryWrapper -> Let's do some research - for now i agree that it is not mentioned in the OpenAPI Spec

PR #151, opened 2026-01-27
Digital Twin Registry

Martin187187:
implements #155

biggest logical changes:
- security doesnt need at least 1 oicd in trustlist per default anymore
- security can work with context parameter now (didnt work before)
- added debug options in configuration (print runtime and queries of descriptor objects)
- changed QL to use inlined EXISTS expressions instead of CTE's
- fixed Discovery Service so it stores specificassetIds
- discovery service can be integrated to the registry now (there is an option)
- new DTR component that combines Discovery and Registry
- swagger for DTR
- Basic Discovery QL
- added createdAt fields in QL
- customizable endpoints for DTR (super minimal addition was needed, ~50 lines of code)
- removed discovery db schema
- DTR injects Edc-Bpn header value into the claim layer

### DIFF
--- a/internal/digitaltwinregistry/description_service.go
+++ b/internal/digitaltwinregistry/description_service.go
@@ -33,8 +33,7 @@ import (
 )
 
 const (
-	discoveryProfile = "https://admin-shell.io/aas/API/3/1/DiscoveryServiceSpecification/SSP-001"
-	registryProfile  = "https://admin-shell.io/aas/API/3/1/AssetAdministrationShellRegistryServiceSpecification/SSP-001"
+	profile = "https://basyx.org/aas/API/3/1/AssetAdministrationShellRegistryServiceSpecification/SSP-001"
 )
 
 // DescriptionService provides the combined service description for the Digital Twin Registry.
@@ -49,6 +48,6 @@ func NewDescriptionService() *DescriptionService {
 func (s *DescriptionService) GetDescription(ctx context.Context) (model.ImplResponse, error) {
 	_ = ctx
 	return model.Response(200, model.ServiceDescription{
-		Profiles: []string{registryProfile, discoveryProfile},
+		Profiles: []string{profile},
 	}), nil
 }


FriedJannik (changes requested):
Thank you for this PR - greate work :)
I added some minor remarks

internal/common/security/authorize.go
```diff
@@ -43,8 +43,9 @@ import (
 // Enabled: toggles ABAC enforcement.
 // Model: provides the AccessModel that evaluates authorization rules.
 type ABACSettings struct {
-	Enabled bool
-	Model   *AccessModel
+	Enabled             bool
+	EnableImplicitCasts bool
```
aaronzi: is that setting only relevant in the security context or also for the query language in general?

internal/digitaltwinregistry/integration_tests/docker_compose/security_env/trustlist.json
aaronzi: is this intended to be empty?

aaronzi (changes requested):
Thanks for this very cool PR. I added a few questions and remarks

internal/digitaltwinregistry/integration_tests/docker_compose/security_env/trustlist.json
Martin187187: Yes. I don’t use any tokens in the security setup. I only use anonymous user with edc header. So i don’t need any oidc

cmd/aasregistryservice/config/access_rules/access-rules.json
```diff
@@ -2,122 +2,42 @@
...
+      { "name": "role_attr", "attributes": [ { "CLAIM": "role" } ] },
+      { "name": "member_attr", "attributes": [ { "CLAIM": "role" }, { "CLAIM": "clear" } ] }
     ],
 
     "DEFOBJECTS": [
-      { "name": "description", "objects": [ { "ROUTE": "/description" } ] },
-      { "name": "all_shell_descriptors", "objects": [ {"DESCRIPTOR": "$aasdesc(\"*\")"}] },
-      { "name": "submodel_descriptors", "objects": [ {"ROUTE": "/shell-descriptors/*/submodel-descriptors"}, {"ROUTE": "/shell-descriptors/*/submodel-descriptors/*"}] }
+      { "name": "description", "objects": [ { "ROUTE": "/description" }, { "ROUTE": "/api/v3/description" } ] },
+      { "name": "all_assetlinks", "objects": [ { "ROUTE": "/lookup/shells" }, { "ROUTE": "/lookup/shells/*" }] },
```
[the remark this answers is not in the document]
Martin187187: I removed the unnecessary route object. Having this route object would not make anything unsafe or incorrect.

internal/common/model/grammar/fieldidentifier_processing.go
```diff
@@ -221,17 +221,18 @@ func ResolveFragmentFieldToSQL(field *FragmentStringPattern) ([]ArrayIndexBindin
...
-	ctxSubmodelDescriptor
-	ctxSubmodelDescriptorEndpoint
-)
+const (
+	ctxUnknown resolveContext = iota
+	ctxAASDesc
+	ctxSMDesc
+	ctxSM
+	ctxSME
+	ctxBD
```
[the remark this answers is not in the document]
Martin187187: Yes. I added a Basic Discovery QL support under the hood.

internal/common/security/edc_bpn.go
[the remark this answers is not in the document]
Martin187187: I took extra care that this addition is only happening in the DTR. If you look into the main.go of the DTR you can see `auth.SetupSecurityWithClaimsMiddleware(ctx, cfg, apiRouter, auth.EdcBpnHeaderMiddleware)` that EdcBpnHeaderMiddleware is only used in the DTR.

internal/digitaltwinregistry/integration_tests/docker_compose/security_env/trustlist.json
Martin187187: obvisouly using the edc-bpn header on its own is not secure. But this is how the customer wants it. Ideally no header should be used and the information is inside the token. What is really nice is that if that is going to be updated in the future, it will work out of the box. Only the Header injection has to be removed.

PR #158, opened 2026-02-02
Registry of Registries (RoR) becomes Registry of Infrastructures (RoI)

chris-koo:
Includes:
- renaming to Registry of Infrastructures
- deletion of unnecessary Infrastructure Descriptor attribute
- addition of RoI integration tests to workflow


aaronzi (commented):
Thanks for the fixes and the renaming of the component. I added two remarks for missing empty last lines but thats just a minor detail. The rest looks good to me 👍

PR #203, opened 2026-03-09
update methods

### LINKED ISSUE #195: [Ticket] Update Method Mapping in Security
## Examples:

change 
`{"PUT", "/shell-descriptors/{aasIdentifier}/submodel-descriptors/{submodelIdentifier}", []grammar.RightsEnum{grammar.RightsEnumUPDATE}}`
to 
`{"PUT", "/shell-descriptors/{aasIdentifier}/submodel-descriptors/{submodelIdentifier}", []grammar.RightsEnum{grammar.RightsEnumUPDATE, grammar.RightsEnumCREATE}}`

---
If access rule has only UPDATE and user uses this rule and a PUT endpoint, then he is only allowed to update existing data not create new. Same logic if rule only has CREATE rights then he is only allowed to create new data even with the put endpoint. Right enums inside the list has to be handeld with an OR.

## Entry Points
- https://github.com/eclipse-basyx/basyx-go-components/blob/main/internal/common/security/abac_engine_methods.go

---

## Acceptance Criteria
- [ ] All existing test have to pass
- [ ] all put method mappings are adjusted
- [ ] extra logic to check if data is already there or not to make the correct decisions for UPDATE and CREATE
- [ ] write new integration tests that tests the enhanced logic
- [ ] Enum Rights are ORed logically.

---

## Risks and Assumptions
- **Risks**: it is possible that his change introduces failing testcases. Fix testcases to this new logic
- **Assumptions**: Worklow of the gates has to be changed. if METHOD is allowed can not be decided in the security layer alone anymore....

Martin187187:
closes #195


internal/discoveryservice/security_tests/it_config.json
aaronzi: Please revert the line ending change

PR #259, opened 2026-04-14
fix example

Martin187187:


aaronzi:
#260 fixes the example shells instead

PR #261, opened 2026-04-15
Align Basyx DTR for crosscutting

Martin187187:
This PR alignes the DTR to this document https://github.com/eclipse-tractusx/sldt-digital-twin-registry/blob/main/docs/architecture/6-crosscutting-concepts.md


internal/common/configuration.go
```diff
@@ -208,7 +208,7 @@ type GeneralConfig struct {
 // OIDCProviderConfig contains OpenID Connect authentication provider settings.
 type OIDCProviderConfig struct {
 	Issuer   string   `mapstructure:"issuer" yaml:"issuer" json:"issuer"`       // OIDC issuer URL
-	Audience string   `mapstructure:"audience" yaml:"audience" json:"audience"` // Expected token audience
+	Audience string   `mapstructure:"audience" yaml:"audience" json:"audience"` // Optional expected token audience (skip audience validation if empty)
```
aaronzi: I think you forgot to remove the "expected" behind "Optional"

cmd/digitaltwinregistryservice/config.yaml
```diff
@@ -2,7 +2,7 @@ server:
   port: 5004
   contextPath: ""
   host: 0.0.0.0
-  strictVerification: true
+  strictVerification: false
```
aaronzi, line 5: Why can strict verification not be used here as default?

cmd/digitaltwinregistryservice/config.yaml
```diff
@@ -19,10 +19,11 @@ oidc:
...
-  enabled: false
+  enabled: true
   modelPath: "config/access_rules/access-rules.json"
 
 general:
-  enableImplicitCasts: true
-  enableDescriptorDebug: false
-  supportsSingularSupplementalSemanticId: false
+  enableImplicitCasts: false
+  enableDescriptorDebug: true
```
aaronzi: What are the reasons for those two settings to be changed?

cmd/digitaltwinregistryservice/config.yaml
```diff
@@ -19,10 +19,11 @@ oidc:
...
+  enableCustomMiddlewareHeaderInjection: false
```
aaronzi, line 28: If this setting is relevant for the DTR, why is it be default disabled?

cmd/digitaltwinregistryservice/main.go
```diff
@@ -135,15 +135,24 @@ func runServer(ctx context.Context, configPath string, databaseSchema string) er
...
+			continue
+		}
 		apiRouter.Method(rt.Method, rt.Pattern, rt.HandlerFunc)
 	}
 	for _, rt := range discoveryCtrl.Routes() {
-		if rt.Method == "POST" && rt.Pattern == "/lookup/shellsByAssetLink" {
+		if (rt.Method == "POST" && rt.Pattern == "/lookup/shellsByAssetLink") || (rt.Method == "GET" && rt.Pattern == "/lookup/shells") {
 			apiRouter.With(digitaltwinregistry.CreatedAfterMiddleware).Method(rt.Method, rt.Pattern, rt.HandlerFunc)
 			continue
 		}
```
aaronzi, line 158: Why are those three routes added with an additional middleware layer? What does this `CreatedAfterMiddleware`layer do?

cmd/digitaltwinregistryservice/openapi.yaml
```diff
@@ -5,9 +5,12 @@ info:
     Combined API for Asset Administration Shell Registry Service and
     Asset Administration Shell Basic Discovery Service.
 
-    Edc-Bpn header: Optional header used in the query language for access control
-    and filtering. If the end user can change this header, the access control is
-    not secure because it is not derived from the token.
+    Edc-Bpn header: Optional header used in the query language for access control
+    and filtering. If the end user can change this header, the access control is
+    not secure because it is not derived from the token. Header injection from
+    this middleware is disabled by default and can be enabled via
```
aaronzi: Why is it disabled by default (same question as above in the example)

cmd/digitaltwinregistryservice/openapi.yaml
```diff
@@ -5,9 +5,12 @@ info:
...
 
-    Edc-Bpn header: Optional header used in the query language for access control
-    and filtering. If the end user can change this header, the access control is
-    not secure because it is not derived from the token.
+    Edc-Bpn header: Optional header used in the query language for access control
+    and filtering. If the end user can change this header, the access control is
+    not secure because it is not derived from the token. Header injection from
+    this middleware is disabled by default and can be enabled via
+    general.enableCustomMiddlewareHeaderInjection (env:
+    GENERAL_ENABLECUSTOMMIDDLEWAREHEADERINJECTION).
```
aaronzi: I'm not sure if this type of description belongs an an OpenAPI file because that is application specific

internal/aasregistry/benchmark_results/insert_data.py
```diff
@@ -7,7 +7,7 @@
 URL = "http://localhost:5004/shell-descriptors"
 JSON_FILE = "bodies/testbench.json"  # path to your file
 INSERT_COUNT = 30000  # Change number if you want more inserts
-INSERT_LIST_ONCE = True  # If JSON file is a list, insert each element once
+INSERT_LIST_ONCE = False  # If JSON file is a list, insert each element once
```
aaronzi: This looks like an unrelated change

internal/common/descriptors/AASDescriptorCreatedAtContext.go
aaronzi, line 1: This is only applied in the DTR, correct?

internal/common/descriptors/AssetAdminShellDescriptorHandler.go
```diff
@@ -373,6 +373,7 @@ func buildListAssetAdministrationShellDescriptorsQuery(
 		pageAlias = "aas_page"
 		dataAlias = "aas_list_data"
 	)
+	includeCreatedAt := includeAASDescriptorCreatedAtFromContext(ctx)
```
aaronzi, line 376: Where does this ctx comes from? I guess this determines that the createdAt is only active for the DTR, correct?

internal/common/descriptors/AssetAdminShellDescriptorHandler.go
```diff
@@ -391,6 +395,24 @@ func buildListAssetAdministrationShellDescriptorsQuery(
...
+		goqu.L("?::text", common.TDescriptorPayload.Col(common.ColDisplayNamePayload)).As("raw_displayname_payload"),
+		goqu.L("?::text", common.TDescriptorPayload.Col(common.ColDescriptionPayload)).As("raw_description_payload"),
+		common.TAASDescriptor.Col(common.ColAASID).As("sort_aas_id"),
+	}
+	if includeCreatedAt {
+		innerSelectExpressions = append(innerSelectExpressions, common.TAASDescriptor.Col(common.ColCreatedAt).As("c6"))
+	} else {
+		innerSelectExpressions = append(innerSelectExpressions, goqu.L("NULL").As("c6"))
+	}
+
```
aaronzi, line 415: What is this inner select expression used for? Is this only relevant for the DTR or the AAS Registry in general?

internal/common/descriptors/AssetAdminShellDescriptorHandler.go
```diff
@@ -567,6 +584,7 @@ func listAssetAdministrationShellDescriptors(
 			&r.GlobalAssetID,
 			&r.IDShort,
 			&r.IDStr,
+			&r.CreatedAt,
```
aaronzi, line 596: Is that the same createdAt that is included in the AssetInformation coming in the AAS V3.2 or is this a backend only database column? Is this also present for the "normal" AAS Registry?

internal/common/descriptors/AssetAdminShellDescriptorHandler.go
```diff
@@ -696,6 +715,14 @@ func listAssetAdministrationShellDescriptors(
...
 }
 
+func nullTimeToPtr(nt sql.NullTime) *time.Time {
+	if !nt.Valid {
+		return nil
+	}
+	t := nt.Time
+	return &t
+}
+
```
aaronzi, line 734: Maybe extract this helper function so that it can be reused. It will probably needed for aas repo and submodel repo in v3.2 as well

internal/common/descriptors/ReadSpecificAssetId.go
```diff
@@ -140,10 +141,14 @@ func ReadSpecificAssetIDsByDescriptorIDs(
...
 		).
+		LeftJoin(
+			externalSubjectReferenceKeyAlias,
+			goqu.On(externalSubjectReferenceKeyAlias.Col(common.ColReferenceID).Eq(externalSubjectReferenceAlias.Col(common.ColID))),
+		).
 		LeftJoin(
 			specificAssetIDPayloadAlias,
 			goqu.On(specificAssetIDPayloadAlias.Col(common.ColSpecificAssetID).Eq(specificAssetIDAlias.Col(common.ColID))),
-		).Select(append([]interface{}{
+		).Distinct().Select(append([]interface{}{
```
aaronzi, line 151: Is this distinct needed because of the newly introduced left join? Does this impact performance? Again, I guess this is just a feature of the DTR, right. Is it only active in that scenario then?

I will probably ask this question multiple times because a lot of code changes where made in the common files that are not specific to the DTR and could potentially affect the behavior of the standardized components

internal/common/model/model_asset_administration_shell_descriptor.go
```diff
@@ -61,6 +62,8 @@ type AssetAdministrationShellDescriptor struct {
 	SpecificAssetIds []types.ISpecificAssetID `json:"specificAssetIds,omitempty"`
 
 	SubmodelDescriptors []SubmodelDescriptor `json:"submodelDescriptors,omitempty"`
+
+	CreatedAt *time.Time `json:"createdAt,omitempty"`
```
aaronzi, line 66: As mentioned above, that field does not exist in descriptors or identifiables yet. Therefore this should not be able to be set in the "normal" Descriptors

internal/common/model/transactionObjects.go
```diff
@@ -552,6 +552,7 @@ type AssetAdministrationShellDescriptorRow struct {
 	GlobalAssetID             sql.NullString
 	IDShort                   sql.NullString
 	IDStr                     string
+	CreatedAt                 sql.NullTime
```
aaronzi, line 555: See comments above regarding `createdAt`

internal/common/testenv/jsonsuite.go
```diff
@@ -492,13 +492,28 @@ func normalizeJSON(input []byte) (string, error) {
...
+		for _, nested := range casted {
+			stripDynamicCreatedAt(nested)
+		}
+	case []any:
+		for _, nested := range casted {
+			stripDynamicCreatedAt(nested)
+		}
+	}
+}
+
```
aaronzi, line 516: Is that the function that is called in the "normal" AAS Registry so that the createdAt is not included there?

I'm not sure if that is a good idea to use a recursive function here. Maybe an approach using a stack would be better here for performance reasons.

internal/digitaltwinregistry/integration_tests/expected/lookup/emptyId.json
aaronzi, line 1: Please double check all your added/edited json files for missing empty last lines

internal/digitaltwinregistry/integration_tests/expected/public_paged_result.json
aaronzi, line 1: Are there any integration tests that include the `createdAt`?

internal/digitaltwinregistry/integration_tests/digitaltwinregistry_integration_test.go
```diff
@@ -98,5 +99,10 @@ func TestIntegration(t *testing.T) {
 				deleteAllDescriptors(t, runner, stepNumber, step.Headers)
 			},
 		},
+		TokenProvider: testenv.NewPasswordGrantTokenProvider(
+			"http://localhost:8080/realms/basyx/protocol/openid-connect/token",
+			"basyx-ui",
+			10*time.Second,
+		),
```
aaronzi, line 106: Why is that here in the integration tests? Was this supposed to be committed?

internal/digitaltwinregistry/tractus-x-integrationtests/docker_compose/docker_compose.yml
```diff
@@ -21,21 +21,22 @@ services:
...
+      - POSTGRES_HOST=db
       - POSTGRES_PORT=5432
       - POSTGRES_USER=admin
       - POSTGRES_PASSWORD=admin123
       - POSTGRES_DBNAME=basyxTestDB
       - POSTGRES_MAXOPENCONNECTIONS=500
       - POSTGRES_MAXIDLECONNECTIONS=500
       - POSTGRES_CONNMAXLIFETIMEMINUTES=5
-      - ABAC_ENABLED=true
+      - ABAC_ENABLED=false
```
aaronzi, line 35: Is ABAC not relevant for the Tractus-X tests? The security aspect are now verified in the newly added integration test setup, right?

internal/digitaltwinregistry/tractus-x-integrationtests/digitaltwinregistry_tractusx_integration_test.go
```diff
@@ -519,6 +519,36 @@ func normalizeJSON(input []byte) (string, error) {
...
+		delete(casted, "createdAt")
+		for _, nested := range casted {
+			stripDynamicCreatedAt(nested)
+		}
+	case []any:
+		for _, nested := range casted {
+			stripDynamicCreatedAt(nested)
+		}
+	}
+}
```
aaronzi, line 550: Redundant code (see comment for extracting this logic in one of the comments above)

internal/digitaltwinregistry/tractus-x-integrationtests/digitaltwinregistry_tractusx_integration_test.go
```diff
@@ -519,6 +519,36 @@ func normalizeJSON(input []byte) (string, error) {
...
+
+	stripDynamicCreatedAt(parsed)
+
+	normalized, err := json.Marshal(parsed)
+	if err != nil {
+		return "", err
+	}
+
+	return string(normalized), nil
+}
```
aaronzi, line 536: Maybe also extract this into a common helper if it will be used elsewhere too in the future

internal/digitaltwinregistry/custom_discovery_service.go
aaronzi, line 1: I think I need a small use case description for what those two new function below are used

aaronzi (commented):
Thanks for the further implementations for the DTR. I added a few questions and remarks below. Please have a look at them.

cmd/digitaltwinregistryservice/config.yaml
Martin187187: Data used in Catena-X does not comply with certain constraints, so I propose keeping this set to false.

cmd/digitaltwinregistryservice/main.go
Martin187187: these 3 methods+endpoints need the additional create query parameter. the middleware injects this

internal/aasregistry/benchmark_results/insert_data.py
Martin187187: This is not unrelated. I use this script for first benchmark impressions. The change is not necessary, as I use this for manual testing only

internal/common/descriptors/AASDescriptorCreatedAtContext.go
Martin187187: correct

internal/common/descriptors/AssetAdminShellDescriptorHandler.go
Martin187187: yes.

internal/common/descriptors/AssetAdminShellDescriptorHandler.go
Martin187187: this is the optional createat parameter which is only requested for the dtr.

internal/common/descriptors/ReadSpecificAssetId.go
Martin187187: this is needed for the discovery integration! So it is not dtr specific. If a user creates multiple shell descriptors via registry endpoint and discovery integration is enabled and shells have the same globalAssetId value then this would create duplicate data

Martin187187: Distinct acts on the fetched data after the limit and filter. so it doesnt affect scalability. so for consistency I would leave it there.

internal/common/model/model_asset_administration_shell_descriptor.go
Martin187187: I dont understand this comment This is actually what is happening. Create At data is ONLY atteched when dtr is used. It is not possible that data is set by the normal registry.

internal/common/testenv/jsonsuite.go
Martin187187: No! Normal AAS Registry does not get createAT data by SQL design. So there is no data to be removed. This is a text function and only used by tests.

Martin187187: This is used in the tests because you can not test the value of createAt in the testcase.

internal/digitaltwinregistry/integration_tests/expected/lookup/emptyId.json
Martin187187: Why is the format so important for test data?

internal/digitaltwinregistry/integration_tests/expected/public_paged_result.json
Martin187187: No. You would need to create hacky tests for createAt (for example test time ranges). I could add this check, but exact checks would be much tricker

internal/digitaltwinregistry/integration_tests/digitaltwinregistry_integration_test.go
Martin187187: Of course this is needed as I included keycloak token requests in the testcases..

internal/digitaltwinregistry/tractus-x-integrationtests/docker_compose/docker_compose.yml
Martin187187: no security apsects are tests in these integration tests. And yes they are tested in the other integration test suite

cmd/digitaltwinregistryservice/config.yaml
Martin187187: for testing purposes. i reverted it.

cmd/digitaltwinregistryservice/config.yaml
Martin187187: enableCustomMiddlewareHeaderInjection is a security risk for users that dont know what they are doing

cmd/digitaltwinregistryservice/openapi.yaml
Martin187187: This should be disabled by default, as enabling it poses a security risk. If someone uses this component without explicitly configuring it, similar to how it is currently used in Catena-X, it could introduce a vulnerability. Therefore, I recommend keeping this set to false by default to ensure safe usage.

cmd/digitaltwinregistryservice/openapi.yaml
Martin187187: i removed some parts of the description but the header can be set in the swagger ui aswell. so yes this information is relevant

internal/common/descriptors/AssetAdminShellDescriptorHandler.go
Martin187187: only database data. no business data. not related to 3.2.

internal/digitaltwinregistry/integration_tests/digitaltwinregistry_integration_test.go
aaronzi: I was wondering about the basyx-ui part

internal/digitaltwinregistry/integration_tests/expected/lookup/emptyId.json
aaronzi: It's more about avoiding potential issues and not about technical requirements. But there is a reason that GitHub shows missing empty last lines in files: https://dan-niles.medium.com/why-github-prs-shows-that-red-circle-understanding-newlines-at-end-of-files-8a4869821694

So we should try to follow this when possible

internal/common/model/model_asset_administration_shell_descriptor.go
aaronzi: Ok, if that's the case then everything is fine. I was just wondering because it was added to the model class that is used by the standardized AAS Registry as well.

Martin187187: i will change the name to DBCreatedAt

Martin187187: should be rejected as input parameter

internal/common/configuration.go
Martin187187: Done

internal/digitaltwinregistry/custom_discovery_service.go
Martin187187: I added a description.

FriedJannik (approved):

aaronzi (approved):

Issue #239, opened 2026-04-07
[Ticket] AAS Upload + Download


aaronzi:
## User Story
> **As a** BaSyx user,  
> **I want** a `/serialization` (download) and `/upload` endpoint in the AAS Environment service,  
> **so that** I can import and export AAS Environment payloads as AASX, JSON, and XML.

---

## Rules
- **Scope anchor**: The base AAS Environment service is a prerequisite and out of scope for this ticket.
- **What**: Implement endpoint logic for `/serialization` and `/upload` for the combined AAS Environment component.
- **What**: Add both endpoints to the combined `openapi.yaml` of the AAS Environment service.
- **What**: Endpoint paths in this ticket are `/serialization` and `/upload`; service prefixing (for example `/api/v3`) is handled by the base/context path configuration.
- **What**: `/upload` must consume `multipart/form-data` and require/forward the MIME type provided in the request for the uploaded file part.
- **What**: Supported upload file types are:
	- Plain Environment file (single JSON or XML file containing AAS, Submodels, and ConceptDescriptions)
	- AASX package file (ZIP container with an Environment payload and optional supplementary files such as File submodel element attachments and AAS thumbnail files)
- **What**: Supported upload media types are:
	- Plain Environment JSON: `application/json`
	- Plain Environment XML: `application/xml`
	- AASX with XML Environment body: `application/aasx+xml` and `application/asset-administration-shell+xml`
	- AASX with JSON Environment body: `application/aasx+json` and `application/asset-administration-shell+json`
- **What**: For AASX upload, packages containing XML or JSON serialized Environment payloads must be supported, including optional supplementary files.
- **What**: Use `github.com/aas-core-works/aas-package3-golang` for packaging/unpackaging behavior.
- **What**: `/serialization` must provide standardized serialization behavior compatible with existing BaSyx API expectations and IDTA serialization semantics.
- **What**: Supported serialization response file types and media types follow the same distinction as upload:
	- Plain Environment JSON: `application/json`
	- Plain Environment XML: `application/xml`
	- AASX with XML Environment body: `application/aasx+xml` and `application/asset-administration-shell+xml`
	- AASX with JSON Environment body: `application/aasx+json` and `application/asset-administration-shell+json`
- **How**: Preserve compatibility with BaSyx Java behavior for upload and serialization contracts (no breaking changes in request/response semantics required by existing clients).
- **How**: Integration tests must cover successful and failing paths for both endpoints, including format and media type handling.
[...]

PR #271, opened 2026-04-29
Upload aasenv

Martin187187:
Implements the upload part of: https://github.com/eclipse-basyx/basyx-go-components/issues/239

Some design decisions differ from those proposed in the ticket:

- There is no full transaction block for AASX files. The reasoning is that users typically prefer having most of the data successfully written rather than everything being rolled back due to a single error (e.g., an invalid concept description).
- I used replace or post logic. so existing submodels would be replaced for example. This makes sense if a user wants to upload a newer version of the aasx for example.

(Also I did not add security yet)
What would be good:
An asynchronous version of the /upload endpoint with progress and result feedback would be highly beneficial. This would allow us to: 
- Provide continuous feedback on upload progress
- Report which parts failed and which succeeded
- Avoid request timeouts for large uploads


internal/aasenvironment/integration_tests/expected/iese_conceptdescriptions.json
aaronzi, line 1: Please format the json file

internal/aasenvironment/integration_tests/expected/iese_shells.json
aaronzi, line 1: Please format the json file

internal/aasenvironment/integration_tests/expected/iese_submodels.json
aaronzi, line 1: Please format the json file

internal/aasenvironment/integration_tests/expected/productionplan_conceptdescriptions.json
aaronzi, line 1: Please format the json file

internal/aasenvironment/integration_tests/expected/productionplan_shells.json
aaronzi, line 1: Please format the json file

internal/aasenvironment/integration_tests/expected/productionplan_submodels.json
aaronzi, line 1: Please format the json file

sql_examples/server/docker-compose.yml
```diff
@@ -7,13 +7,6 @@ services:
...
-    
-  adminer:
-    image: adminer:5
-    container_name: adminer
-    ports:
-      - "8080:8080"
-    depends_on:
-      - db
\ No newline at end of file
+
```
aaronzi: Missing empty last line

aaronzi (commented):
Thanks for implementing the AAS upload. I just have very minor remarks. If I should fix them myself, please send me a ping

aaronzi:
Actually one question came to mind. I tried to upload the same aasx file again and that worked without any API provided warning or log in the container. How is this currently handled? Are duplicates always automatically ignored?

aaronzi:
Also Copilot review remarks seem to be valid and need to be addressed

cmd/aasenvironmentservice/openapi.yaml
```diff
@@ -1068,6 +1068,32 @@ components:
...
+                  format: binary
+      responses:
+        '200':
+          description: File uploaded successfully
+        '400':
+          description: Bad request
+        '415':
+          description: Unsupported Media Type
+        '500':
+          description: Internal server error
```
aaronzi, line 1167: Please add examples for error responses here like with the standardized ebdpoints

Martin187187:
> Actually one question came to mind. I tried to upload the same aasx file again and that worked without any API provided warning or log in the container. How is this currently handled? Are duplicates always automatically ignored?

I explained that in the description:
--> I used replace or post logic. so existing submodels would be replaced for example. This makes sense if a user wants to upload a newer version of the aasx for example.

aaronzi:
> I explained that in the description:
> --> I used replace or post logic. so existing submodels would be replaced for example. This makes sense if a user wants to upload a newer version of the aasx for example.

You mean in the PR description, right? I missed that. It needs to be part of the OpenAPI Description as well because that's an important info to the user of the API/component

PR #323, opened 2026-05-26
added missing index

### LINKED ISSUE #310: [BUG] DTR creates multiple database connections instead of shared one
**Describe the bug**
DTR creates multiple database connections instead of shared one

See main.go in the DTR service

**Expected behavior**
Should use shared connection like in the AAS Env.

Martin187187:
closes #310 and added missing indexes


FriedJannik (approved):
lgtm. thx

PR #325, opened 2026-05-27
Bump Go 1.25.9 -> 1.26.3 & SNAPSHOT Workflow changes

FriedJannik:
## Description of Changes

This PR bumps Go to 1.26.3 and adapts the Docker SNAPSHOT Workflow.
Images are now not automatically published after merge, instead a PR needs to be labeled with "docker-snapshot" in order for images to be build. Alternatively a maintainer can start the workflow manually in the actions tab-


.github/workflows/go-tests.yml
FriedJannik, line 1: Most likely someone didn't lint this file before - therefore the diff

aaronzi (approved):

PR #326, opened 2026-05-27
Updates Vulnerability Scan Workflow

FriedJannik:


aaronzi (approved):

aaronzi (changes requested):
please also update this file:

https://github.com/eclipse-basyx/basyx-go-components/blob/main/docu/security/SUPPLY_CHAIN_SECURITY.md

There is currently a statement that trivy also runs on push to main

aaronzi (approved):

PR #327, opened 2026-05-28
Update to Metamodel Version 3.2

FriedJannik:
## Description of Changes

Please provide a brief summary of what you have changed in this pull request. Be clear and concise.

## Related Issue

If this pull request is related to an existing issue, please link that issue using the format `#issue_number`.

Describe any specific configuration settings or environment setup required to test these changes effectively. If possible, please provide the BaSyx configuration as .zip file.

## AAS Files Used for Testing

Please provide any AAS files that were used in the testing of these changes. Include a brief descriptions of their relevance to the changes being proposed.

## Additional Information

Include any additional information or context that could be useful for the reviewer. This could include challenges faced, alternative solutions that were considered, etc.

---

Please ensure that you have tested your changes thoroughly before submitting the pull request.


database/patches/1_1_1.sql
aaronzi, line 1: The history tables should be in a different database in the same Postgres so that individual DB user rights can be configured.

internal/common/history/history.go
aaronzi, line 1: for recent changes, what happens if created at and updated from are both set? does updated remove the where condition for created or are the OR or AND combined?

internal/common/history/history.go
```diff
@@ -0,0 +1,709 @@
...
+		return common.NewInternalServerError("HISTORY-GUARD-BUILDUPDATE " + err.Error())
+	}
+	result, err := db.ExecContext(ctx, updateSQL, updateArgs...)
+	if err != nil {
+		return common.NewInternalServerError("HISTORY-GUARD-EXECUPDATE " + err.Error())
+	}
+	affected, err := result.RowsAffected()
+	if err != nil || affected > 0 {
+		return nil
+	}
```
Martin187187: this can swallow errors. is that intended? maybe have two seperate if checks?

pkg/aasregistryapi/api.go
Martin187187, line 1: why only aas registry? why not sm registry aswell?

Martin187187: and why only GetAllAssetAdministrationShellsRecentChanges?

Martin187187 (approved):

pkg/aasregistryapi/api.go
aaronzi: Good question. The spec defines it like this. But I agree, it's strange. I added this to the review findings at the IDTA.

internal/common/history/history.go
aaronzi: Good point, we can add two separate checks here

aaronzi:
A server that was started with audit active should never allow to disable it at a later point in time (or at least log the deactivation)

cmd/conceptdescriptionrepositoryservice/openapi.yaml
```diff
@@ -255,6 +258,38 @@ paths:
...
+      - $ref: https://api.swaggerhub.com/domains/Plattform_i40/Part2-API-Schemas/V3.2.0#/components/parameters/UpdatedFrom
+      - $ref: https://api.swaggerhub.com/domains/Plattform_i40/Part2-API-Schemas/V3.2.0#/components/parameters/Limit
+      - $ref: https://api.swaggerhub.com/domains/Plattform_i40/Part2-API-Schemas/V3.2.0#/components/parameters/Cursor
+      responses:
+        "200":
+          description: Requested list of recently changed Concept Descriptions
+          content:
+            application/json:
+              schema:
+                $ref: https://api.swaggerhub.com/domains/Plattform_i40/Part1-MetaModel-Schemas/V3.2.0#/components/schemas/ConceptDescription
```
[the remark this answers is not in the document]
aaronzi: Problem in the IDTA openapi.yml

Reported in the AAS API v3.2 findings

Martin187187:
as a user i would expect that /$recent-changes sorts recent changes from latest to oldest. Currenlty it is sorted from oldest to latest. I assume that is specified in the standard?

Martin187187:
Error removing Submodel Element!

Status: 500
Message Type: Error
Correlation ID: SMREPO-500-DeleteSubmodelElementByPathSubmodelRepo-InternalServerError-DeleteSubmodelElementByPath
Timestamp: 6/2/2026, 5:21:20 PM
Text: 500 Internal Server Error: SMREPO-DELSMEBPATH-UPDATEPATH-EXEC Failed to execute update path query: pq: duplicate key value violates unique constraint "uq_sibling_idshort" (23505)

How to reproduce:
I added a submodel element list with 50 items that dont have an idShort. Then i randomly started to delete some. 
(dont know if those errors have anything to to with this PR)

Martin187187:
Error creating Submodel Element!

Status: 500
Message Type: Error
Correlation ID: SMREPO-500-PostSubmodelElementSubmodelRepo-InternalServerError-AddSubmodelElement
Timestamp: 6/2/2026, 5:32:54 PM
Text: 500 Internal Server Error: SMREPO-INSSME-INSTYPE-EXECQ pq: invalid input syntax for type json at column 80 (22P02)

To reproduce:
- add AnnotatedRelationshipElement to a submodel

(dont know if those errors have anything to to with this PR)

aaronzi:
> as a user i would expect that /$recent-changes sorts recent changes from latest to oldest. Currenlty it is sorted from oldest to latest. I assume that is specified in the standard?

It's not defined in the spec as far as I can see. But agreed, it makes sense to sort from newest to oldest. Just forgot about that. The DB obviously stores it from oldest to newest in the append-only history tables.

internal/common/history/history.go
aaronzi: Only on of the two can happen at the same time and it makes sense that the latter one replaces the first one

database/patches/1_1_1.sql
aaronzi: We decided to not do this. The external anchor will archive this exact requirement. If users need that, the should use the external anchor.

PR #338, opened 2026-06-04
Adds new history diff mode backend

aaronzi:
## Summary

This PR introduces diff-backed history storage for AAS versioning and auditing.

Instead of storing a complete snapshot for every version, the history backend can now store periodic full snapshots combined with RFC 6902 JSON Patch diffs between them. This significantly reduces storage requirements for frequently changing AAS resources while preserving the ability to reconstruct and verify every historical version.

## Motivation

The initial implementation stored a full serialized snapshot for every version. While simple and robust, this approach can become storage-intensive for large AASs with frequent updates where only small portions of the payload change between versions.

This PR adds a hybrid storage model that combines:

- Periodic full snapshot checkpoints
- RFC 6902 JSON Patch diffs between checkpoints
- Integrity verification for both stored payloads and reconstructed versions
- Backward compatibility with existing snapshot-only history records

The goal is to improve scalability without sacrificing auditability, traceability, or deterministic version reconstruction.

### Diff-backed history storage

- Added support for storing RFC 6902 JSON Patch diffs between versions.
- Introduced configurable snapshot checkpoint intervals via history.fullSnapshotInterval.
- Versions between checkpoints are stored as diffs whenever possible.
- Full snapshots continue to be supported and remain the storage format for checkpoint versions.

### Version reconstruction

- Implemented reconstruction of historical versions from the nearest checkpoint plus subsequent diffs.
- Added integrity verification during reconstruction.
- Historical reads and recent-change feeds transparently return fully reconstructed resources.

### Hashing and auditability

- Added separate hashes for:
  - Reconstructed resource content (content_hash)
  - Persisted storage payload (payload_hash)
- Enables verification of both historical resource state and stored history records.
- Preserves deterministic audit trails across snapshot and diff records.

### Compatibility

- Existing snapshot-only history remains fully supported.
- No migration is required for existing installations.
- Snapshot and diff-based history records can coexist.

### Documentation

- Updated runtime and configuration documentation to describe:
  - Diff-backed storage behavior
  - Checkpoint interval configuration
  - Reconstruction process
  - Operational and scalability considerations

## Benefits

- Reduced storage consumption for frequently changing AAS resources.
- Bounded reconstruction cost through configurable checkpoints.
- Full auditability and integrity verification.
- Backward compatibility with existing history data.
- Foundation for future auditing and compliance features built on AAS v3.2 versioning.


internal/conceptdescriptionrepository/integration_tests/conceptdescriptionrepository_recent_changes_integration_test.go
```diff
@@ -83,6 +86,7 @@ func TestConceptDescriptionRepositoryRecentChanges(t *testing.T) {
 	if status != http.StatusNoContent {
 		t.Fatalf("expected 204, got %d body=%s", status, string(body))
 	}
+	requireConceptDescriptionHistoryPayloadTypes(t, conceptDescriptionID, []string{"snapshot", "diff"})
```
chris-koo: Since the snapshot interval is set to three it could make sense to test a longer history than just two

internal/common/history/history.go
```diff
@@ -27,545 +27,97 @@
...
-		return AuditIdentityMinimal
-	case AuditIdentityExtended:
-		return AuditIdentityExtended
-	default:
-		return AuditIdentityMinimal
-	}
+	return restoreVersionByHistoryID(ctx, tx, table, identifier, historyID)
 }
 
 // SnapshotByDate returns the snapshot that was valid for identifier at the requested instant.
```
chris-koo: It could be beneficial to have more elaborated comments describing the exported functions

chris-koo (approved):

PR #345, opened 2026-06-05
Adds WORM evidence storage for history events and snapshots

aaronzi:
## Summary

Adds WORM-compatible history evidence support for AAS v3.2 history.

The default integrity architecture is now:

`PostgreSQL history tables -> hash chain -> signed/unsigned manifest -> S3-compatible WORM object storage`

This keeps the existing `$history` and `$recent-changes` HTTP APIs unchanged, while adding external evidence artifacts for tamper detection, verification, and bounded recovery.

## Changes

- Added `EvidenceStore` abstraction for immutable evidence artifacts.
- Added S3-compatible evidence store using AWS SDK for Go v2.
- Added fail-closed synchronous `history_event` artifact writes during history append:
  - object-store write happens before PostgreSQL commit
  - PostgreSQL transaction rolls back if evidence storage fails
  - receipt is stored in PostgreSQL only after the object write succeeds
- Added per-row WORM `history_event` artifacts containing:
  - history payload, snapshot or diff
  - row hash metadata
  - audit/source fields where currently available
  - `effective_diff` for attribution, even when the stored recovery payload is a full snapshot
- Added signed/unsigned history manifests for history ranges.
- Added PostgreSQL catalog tables for evidence manifests and artifacts via `1_1_3.sql`.
- Added verifier/publisher CLI: `cmd/historyevidenceverifier`.
- Added S3/MinIO example configuration to `BaSyxHistoryAuditGuardedExample`.
- Kept `IntegrityAnchor` / `external_anchor` as optional future extension point for immudb, Rekor, Trillian, or timestamping.
- Updated developer/user documentation for evidence storage, verification, recovery limits, and operational caveats.

## Important Behavior

When evidence storage is enabled, successful API writes require both PostgreSQL and WORM evidence storage to succeed. If the S3-compatible backend is unavailable, rejects retention metadata, or times out, the mutation fails and PostgreSQL rolls back.

Recovery from WORM evidence is bounded by `history.fullSnapshotInterval`:
- `1`: every row is recoverable as a full WORM snapshot
- `>1`: recovery starts from the latest WORM snapshot and replays WORM-stored diffs

The `effective_diff` field prevents attribution confusion when a full snapshot checkpoint is written for a small user change.

## Security / Compliance Notes

This feature strengthens integrity protection, traceability, auditability, recovery evidence, and unauthorized modification detection. It does not by itself make a deployment NIS2-compliant.

Production deployments still need correctly operated WORM storage, restricted IAM, retention policy, key management, monitoring, audit logs, backups, and operational procedures.

MinIO is included only as a local/test S3-compatible Object Lock backend. Production targets should use an operated S3-compatible WORM service, for example AWS S3 Object Lock.


Martin187187 (commented):
is audit information intentionally not added yet?
i get this in minio: "audit":{"actor_issuer":"","actor_subject":"","authorization_result":"","client_id":"","correlation_id":"","endpoint":"","http_method":"","matched_rule_id":"","operation":"","policy_id":"","request_id":"","source_ip":"","user_agent":""}

Martin187187:
image
Deleted Objects are shown no matter if "Show deleted objects" is ticked or not. Maybe this can be configured when requesting the histroy update to minio?

aaronzi:
> is audit information intentionally not added yet?

Yes, that will be introduced with the next PR

aaronzi:
> Deleted Objects are shown no matter if "Show deleted objects" is ticked or not. Maybe this can be configured when requesting the histroy update to minio?

Deleted here means deleted from the WORM storage, not deleted in Terms of an AAS was deleted. The WORM storage is also append only. You can't delete stuff from there. And even if you do, you can not fully delete it and will always see it if you check this checkbox.

Btw, history ist not fetched from the WORM storage. It always comes from the PostgreDB. The WORM storage is a recovery and traceability storage, not for providing the data for the /$history endpoints

Martin187187:
Error updating Property "PcfCO2eq"!

Status: 500
Message Type: Error
Correlation ID: SMREPO-500-PatchSubmodelElementByPathValueOnlySubmodelRepo-InternalServerError-UpdateSubmodelElementValueOnly
Timestamp: 6/8/2026, 11:06:04 AM
Text: 500 Internal Server Error: HISTORY-EVIDENCE-APPEND-PUTARTIFACT HISTORY-EVIDENCE-S3-PUTOBJECT operation error S3: PutObject, https response error StatusCode: 0, RequestID: , HostID: , request send failed, Put "http://minio:9000/basyx-history-evidence/history-evidence/history-events/submodel_history/urn%3Afraunhofer%3Aiese%3Adte%3Asm%3Acarbonfootprint%3Adrivemotor-dm3000%3A001/14-1c8c3c43d01320f88517293cdfad5bd290c10be32a4e4fe834e33e5d60a7b909.json?x-id=PutObject": dial tcp: lookup minio on 10.89.5.1:53: no such host

I get thiss error when i start the backend wihtout minio running and trying to change a value in a submodel. Maybe 503 Service Unavailable is more fitting?

internal/common/history/evidence_store_s3.go
```diff
@@ -0,0 +1,342 @@
...
+		Metadata:    cleanS3Metadata(receipt.Metadata),
+	}
+	applyS3Retention(input, receipt)
+	output, err := store.client.PutObject(ctx, input)
+	if err != nil {
+		return nil, fmt.Errorf("HISTORY-EVIDENCE-S3-PUTOBJECT %w", err)
+	}
+	if output.VersionId != nil {
+		receipt.Reference.VersionID = strings.TrimSpace(*output.VersionId)
+	}
```
Martin187187, line 178: Maybe enforce a versionId aswell? This could go really wrong if config is changed and s3 retuns versions suddendly and same object is written again. The "old version" that didnt had a version is basically overwritten. This is a extrem edge case. I dont know if that is even relevant

internal/common/history/evidence_verifier.go
```diff
@@ -0,0 +1,510 @@
...
+	if receipt.ContentHash != "" && receipt.ContentHash != candidate.ContentHash {
+		report.addFinding(VerificationSeverityError, "HISTORY-EVIDENCE-VERIFY-EVENTCONTENTHASH", "history_event receipt content_hash does not match PostgreSQL row", candidate.Identifier, candidate.HistoryID)
+	}
+	if receipt.Receipt.RetentionMode == "" || receipt.Receipt.RetainUntil == nil {
+		report.addFinding(VerificationSeverityError, "HISTORY-EVIDENCE-VERIFY-EVENTRETENTION", "history_event receipt is missing WORM retention metadata", candidate.Identifier, candidate.HistoryID)
+	}
+	if options.EvidenceStore == nil {
+		return
+	}
+	if _, err := options.EvidenceStore.VerifyArtifact(ctx, receipt.Receipt.Reference, receipt.Receipt.SHA256); err != nil {
```
Martin187187, line 390: This only verifies object content. The retention check currently trusts PostgreSQL receipt metadata. For WORM assurance, verification should query S3/MinIO for the actual object retention/legal-hold state of this version and compare it to the DB receipt. I understand that this is not easy to implement but the verification should happen on the source of truth that is minio in my opinion.

internal/common/history/evidence_verifier.go
```diff
@@ -0,0 +1,510 @@
...
+		return err
+	}
+	if _, err = restoreVersionChainRows(rows); err != nil {
+		return err
+	}
+	return nil
+}
+
+func verifyManifestArtifact(ctx context.Context, options VerifyHistoryRangeOptions, report *HistoryEvidenceVerificationReport) {
+	if options.EvidenceStore == nil || strings.TrimSpace(options.ManifestArtifactRef.ObjectKey) == "" || strings.TrimSpace(options.ManifestArtifactHash) == "" {
```
Martin187187: this can silently skip verification even if user tried to add a object-key if manifest-sha256 is missing for example.
so 
```
go run ./cmd/historyevidenceverifier `
  -table aas_history `
  -from 1 `
  -to 10 `
  -manifest-object-key history-evidence/manifests/aas_history/1-10.json
```

check would be skipped because -manifest-sha256 is missing. The user might not expect the silent skip in verification.

maybe dont skip verification if one of them is missing? This is a nitpick as well.

internal/common/history/evidence_store_s3.go
aaronzi: Can be. I improved that based on your comment

internal/common/history/evidence_verifier.go
aaronzi: True. Verifier now checks the actual retention and legal-hold state through GetObjectRetention / GetObjectLegalHold

internal/common/history/evidence_verifier.go
aaronzi: Good point. Manifest artifact verification now no longer silently skips partial options. The CLI now rejects -manifest-object-key without -manifest-sha256, and the verifier reports partial config as a finding.

PR #351, opened 2026-06-09
Enhance ABAC with deterministic matched rule IDs for audit enrichment

### LINKED ISSUE #350: [Ticket] Deterministic ABAC Rule IDs For Audit Evidence
## Summary

Add deterministic ABAC matched rule identifiers so history audit rows and WORM evidence artifacts can record which configured ABAC rule allowed a request.

## Background

The current history audit enrichment can store `authorization_result` and an ABAC `policy_id` hash. `policy_id` identifies the complete configured access policy, but `matched_rule_id` is still empty because the ABAC evaluator does not expose a stable identifier for the specific rule that matched.

Request and correlation IDs for HTTP audit rows are intentionally not generated by BaSyx audit middleware. Clients, API gateways, or trusted reverse proxies should set `X-Request-ID` and `X-Correlation-ID`; BaSyx copies those headers when present.

## Goals

- Extend ABAC evaluation to return a deterministic matched rule id for allow decisions.
- Store that rule id in `AuthorizationDecision.MatchedRuleID`.
- Preserve the existing `policy_id` behavior as the hash of the complete ABAC policy.
- Keep `policy_id + matched_rule_id` useful for audit traceability without storing policy contents in history rows.

## Proposed Rule ID Format

Use a deterministic, non-secret identifier derived from the matched rule position and canonical rule content, for example:

```text
rule:<index>:<sha256-prefix>
```

The exact format can be adjusted during implementation, but it should be stable for unchanged policy files and deterministic across service restarts.

## Constraints

- Do not change existing public AAS/Submodel/ConceptDescription history API response shapes.
- Do not store full ABAC policy documents in history rows or WORM artifacts.
- Do not expose bearer tokens, raw claims, or other secrets.
- Keep behavior backward compatible when no matched rule id is available.

## Acceptance Criteria

- ABAC allow decisions expose a deterministic matched rule id.
- `history.auditIdentityMode=extended` copies the matched rule id into PostgreSQL history rows when available.
- WORM `history_event` artifacts contain the same `matched_rule_id` as the corresponding PostgreSQL history row.
- Anonymous, ABAC-disabled, and route-not-found/pass-through requests remain valid with empty `matched_rule_id`.
- Existing configs remain readable and do not require adding explicit rule ids.

## Tests

- Unit test deterministic rule id generation for unchanged policy input.
- Unit test that changing the matched rule content changes the hash portion.
- Unit test that matched rule ids are populated on ABAC allow decisions.
- Audit middleware/history tests for propagation into PostgreSQL row metadata.
- WORM event artifact tests for matching `matched_rule_id` propagation.

aaronzi:
## Summary

Adds deterministic ABAC matched rule identifiers so history audit rows and WORM `history_event` artifacts can record which configured ABAC allow rule(s) authorized a request.

## What Changed

- Added shared canonical JSON hashing helpers in `internal/common`.
- Kept existing `history.CanonicalJSON*` APIs as wrappers to avoid churn.
- Precomputes deterministic rule IDs during ABAC rule materialization:
  - format: `rule:<1-based-index>:<first16-sha256-hex>`
  - hash input: canonical JSON of the configured rule entry
- Added detailed ABAC evaluation metadata while preserving existing `AuthorizeWithFilter*` signatures.
- Stores matched rule IDs in `AuthorizationDecision.MatchedRuleID`.
- Supports multiple matched allow rules by storing comma-separated IDs in configured rule order.
- Propagates matched rule IDs through extended history audit metadata into PostgreSQL rows and WORM artifacts.
- Updated the guarded history example docs to describe the new `matched_rule_id` behavior.

Closes #350

### DIFF
--- a/internal/common/security/abac_engine.go
+++ b/internal/common/security/abac_engine.go
@@ -153,42 +153,17 @@ type AuthorizationEvaluation struct {
 //
 //nolint:revive // i will refactor this function at some point
 func (m *AccessModel) AuthorizeWithFilter(in EvalInput) (bool, DecisionCode, *QueryFilter) {
-	return m.AuthorizeWithFilterWithOptions(in, grammar.DefaultSimplifyOptions())
-}
-
-// AuthorizeWithFilterWithOptions behaves like AuthorizeWithFilter but allows callers
-// to control backend simplification behavior (e.g., implicit casts).
-func (m *AccessModel) AuthorizeWithFilterWithOptions(in EvalInput, opts grammar.SimplifyOptions) (bool, DecisionCode, *QueryFilter) {
-	result := m.AuthorizeWithFilterDetailedWithOptions(in, opts)
+	result := m.AuthorizeWithFilterWithOptions(in, grammar.DefaultSimplifyOptions())
 	return result.Allowed, result.Reason, result.QueryFilter
 }
 
-// AuthorizeWithFilterDetailed evaluates the request and returns audit-friendly metadata.
+// AuthorizeWithFilterWithOptions evaluates the request and returns audit-friendly metadata.
 //
-// The evaluation semantics are identical to AuthorizeWithFilter. In addition to
-// the allow decision and optional query filter, the returned value contains the
-// deterministic matched rule identifiers used by history audit enrichment.
-//
-// Parameters:
-//   - in: Request method, path, and claims used by the ABAC evaluator.
-//
-// Returns:
-//   - AuthorizationEvaluation: Access decision, decision reason, optional query
-//     filter, and comma-separated matched allow rule IDs.
-//
-// Example:
-//
-//	result := model.AuthorizeWithFilterDetailed(input)
-//	if !result.Allowed {
-//		return result.Reason
-//	}
-//	audit.MatchedRuleID = result.MatchedRuleID
-func (m *AccessModel) AuthorizeWithFilterDetailed(in EvalInput) AuthorizationEvaluation {
-	return m.AuthorizeWithFilterDetailedWithOptions(in, grammar.DefaultSimplifyOptions())
-}
-
-// AuthorizeWithFilterDetailedWithOptions behaves like AuthorizeWithFilterWithOptions
-// and additionally returns the deterministic IDs of all matched allow rules.
+// The evaluation semantics are identical to AuthorizeWithFilter, but callers can
+// control backend simplification behavior, for example implicit casts. In
+// addition to the allow decision and optional query filter, the returned value
+// contains the deterministic matched rule identifiers used by history audit
+// enrichment.
 //
 // Parameters:
 //   - in: Request method, path, and claims used by the ABAC evaluator.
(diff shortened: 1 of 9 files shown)


internal/common/security/abac_engine.go
```diff
@@ -134,18 +158,71 @@ func (m *AccessModel) AuthorizeWithFilter(in EvalInput) (bool, DecisionCode, *Qu
...
+//	result := model.AuthorizeWithFilterDetailedWithOptions(input, opts)
+//	if result.Allowed {
+//		ctx = ContextWithAuthorizationDecision(ctx, AuthorizationDecision{
+//			Result:        string(DecisionAllow),
+//			MatchedRuleID: result.MatchedRuleID,
+//		})
+//	}
+//
+// nolint:revive // This function is the heart of ABAC and is complicated. Sorry cognitive-complexity!
+func (m *AccessModel) AuthorizeWithFilterDetailedWithOptions(in EvalInput, opts grammar.SimplifyOptions) AuthorizationEvaluation {
```
Martin187187: instead of using a wrapper this could be implemented directly into `func (m *AccessModel) AuthorizeWithFilterWithOptions(in EvalInput, opts grammar.SimplifyOptions) (bool, DecisionCode, *QueryFilter)`. So (bool, DecisionCode, *QueryFilter) replaced by AuthorizationEvaluation.
This would make the code a bit more readable in my opinion.

internal/common/security/abac_engine.go
```diff
@@ -122,6 +124,28 @@ const (
 	DecisionRouteNotFound DecisionCode = "ROUTE_NOT_FOUND"
 )
 
+// AuthorizationEvaluation records the detailed outcome of an ABAC evaluation.
+//
+// ABACMiddleware uses this richer result to pass allow metadata to audit
+// middleware while preserving the older AuthorizeWithFilter return contract for
+// callers that only need access and query-filter decisions.
+type AuthorizationEvaluation struct {
```
Martin187187, line 132: Maybe a better name would be AuthorizationEvaluationResult? I think AuthorizationEvaluation is not enough. I think this struct is really usefull for the readability and can replace  (bool, DecisionCode, *QueryFilter) entirely.

internal/common/security/abac_engine_materialization.go
```diff
@@ -56,12 +59,24 @@ func materializeRules(all grammar.AccessRuleModelSchemaJSONAllAccessPermissionRu
 		if err != nil {
 			return nil, fmt.Errorf("rule %d: %w", i+1, err)
 		}
+		mr.id, err = deterministicMatchedRuleID(i+1, r)
```
Martin187187, line 62: Maybe introduce an object called DeterministicRule, which has the hash and the index number? Then we could use a list of these objects instead of a string. What is the advantage of using a concatenated string here?

aaronzi: That's the most pragmatic way right now. Later on when access rules are stored in the DB, this can change

PR #381, opened 2026-06-17
Enhance supplementalSemanticIds querying and add Catena-X setup example

Martin187187:
- added ql support for supplementalSemanticIds (filter+condition)
  - updates schema for QL
  - has migration tests, so it works with old data aswell
- added query endpoint for sm descritpor
- added example for dtr + submodel service setup (catena-x)


FriedJannik (approved):
I have no further remarks

aaronzi (approved):

PR #394, opened 2026-06-20
improve dtr & aas registry inserts massivly

Martin187187:
- reduces network overhead to a minimum (from 11k to 1 for bulk endpoint with 1000 twins)

## Before

Previously, the bulk endpoint processed each descriptor individually:

```text
check whether this ID exists
insert one descriptor and its nested objects
repeat for the next descriptor
```

This resulted in many database operations for every descriptor, including inserts for:

* AAS descriptors
* Submodel descriptors
* Endpoints
* Specific asset IDs
* Semantic references
* Payload rows
* Other related nested data

For large bulk requests, this created substantial database and network overhead.

## After

The new bulk-create flow processes the complete request as a batch:

```text
validate whole request
BEGIN transaction
check all requested AAS IDs at once
reserve all generated database IDs needed
flatten all descriptors into table rows
execute ordered multi-row INSERT statements
read back (security) / write history
COMMIT
```

## How It Works

1. **Validate the complete request**
   The request is checked for invalid or duplicate descriptor IDs before data is written.

2. **Check existing IDs in bulk**
   Instead of querying the database once per descriptor, all requested AAS IDs are checked together.

3. **Reserve database IDs up front**
   IDs for descriptors and nested entities are generated before insertion so that foreign-key relationships can be built correctly.

4. **Flatten descriptors into database rows**
   Each descriptor and its nested objects are converted into rows for the relevant database tables.

5. **Insert rows in dependency order**
   The generated rows are inserted using multi-row SQL statements, ordered so parent records are written before dependent child records.

6. **Keep the operation atomic**
   All inserts run inside a single transaction. If any step fails, the transaction is rolled back and no partial bulk import remains.

7. **Write history after persistence**
   Successfully created descriptors are read back and passed to the history mechanism.

## Result

The main improvement is that bulk creation no longer performs a full persistence workflow separately for every descriptor.

Instead, it performs batched validation, batched ID checks, batched row generation, and multi-row inserts. This significantly reduces write overhead for large bulk imports while preserving all-or-nothing transactional behavior.

> Note: The write operations are batched, but the full request may still include additional database interactions for ID reservation, validation, and history handling.


internal/common/descriptors/postgres_bulk_create_batch.go
```diff
@@ -0,0 +1,539 @@
...
+	"fmt"
+
+	"github.com/FriedJannik/aas-go-sdk/types"
+	"github.com/doug-martin/goqu/v9"
+	"github.com/doug-martin/goqu/v9/exp"
+	"github.com/eclipse-basyx/basyx-go-components/internal/common"
+	"github.com/eclipse-basyx/basyx-go-components/internal/common/model"
+)
+
+const bulkInsertRowLimit = 1000
```
Martin187187: This could(should) be an config variable maybe...

internal/aasregistry/api/api_asset_administration_shell_registry_api_service.go
aaronzi, line 1: That's an unrelated change for a "problem" that was missed before, correct?

internal/aasregistry/api/bulk_atomic_operations_test.go
```diff
@@ -1,65 +1,46 @@
...
-* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
-* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
-* NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
-* LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
-* OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
-* WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
-*
 * SPDX-License-Identifier: MIT
 ******************************************************************************/
-// Author: Aaron Zielstorff ( Fraunhofer IESE )
```
aaronzi, line 25: Please revert those changes

internal/aasregistry/api/bulk_atomic_operations.go
```diff
@@ -43,13 +43,69 @@ func (s *AssetAdministrationShellRegistryAPIAPIService) ExecuteBulkCreateAtomic(
 	ctx context.Context,
 	descriptors []model.AssetAdministrationShellDescriptor,
 ) asyncbulk.OperationResult {
-	return s.executeAtomicAASDescriptorBulk(
-		ctx,
-		descriptors,
+	if len(descriptors) == 0 {
+		return successfulAtomicResult(0)
+	}
```
aaronzi, line 48: Should sending no descriptors be a successful request? Was it like this before?

internal/aasregistry/api/bulk_atomic_operations_test.go
aaronzi, line 1: Is there any reason why you removed test cases here instead of just adding your new test cases?

go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	github.com/coreos/go-oidc/v3 v3.18.0
 	github.com/doug-martin/goqu/v9 v9.19.0
 	github.com/go-chi/cors v1.2.2
+	github.com/jackc/pgx/v5 v5.8.0
```
aaronzi, line 18: Why are we introducing a second postgres driver? Everywhere else we use `https://github.com/lib/pq`

internal/submodelrepository/persistence/submodel_write_database.go
aaronzi, line 1: Refactor in a separate PR instead of this one?

internal/submodelrepository/persistence/submodelElements/submodel_element_database_utils.go
aaronzi, line 1: Thats just a cosmetic change/refactor, correct? Maybe separate PR?

internal/common/WriteContextReference.go
aaronzi, line 1: Unrelated?

internal/common/postgres_batch_test.go
```diff
@@ -0,0 +1,50 @@
+/*******************************************************************************
+* Copyright (C) 2026 the Eclipse BaSyx Authors and Fraunhofer IESE
+*
+* SPDX-License-Identifier: MIT
+******************************************************************************/
```
aaronzi, line 24: Fix license header

internal/common/postgres_batch_descriptor.go
```diff
@@ -0,0 +1,163 @@
+/*******************************************************************************
+* Copyright (C) 2026 the Eclipse BaSyx Authors and Fraunhofer IESE
+*
+* SPDX-License-Identifier: MIT
+******************************************************************************/
```
aaronzi, line 24: Fix license header

internal/common/postgres_batch.go
```diff
@@ -0,0 +1,146 @@
+/*******************************************************************************
+* Copyright (C) 2026 the Eclipse BaSyx Authors and Fraunhofer IESE
+*
+* SPDX-License-Identifier: MIT
+******************************************************************************/
```
aaronzi, line 24: Fix license header

internal/common/error_handler_test.go
```diff
@@ -30,6 +30,9 @@ import (
 	"fmt"
 	"net/http"
 	"testing"
+
+	"github.com/jackc/pgx/v5/pgconn"
+	"github.com/lib/pq"
```
aaronzi: Why do we need to support both at the same time? Why not fully switch to pgx? Or alternatively why not stay on pq only?

internal/common/descriptors/postgres_create_batch_test.go
```diff
@@ -0,0 +1,68 @@
+/*******************************************************************************
+* Copyright (C) 2026 the Eclipse BaSyx Authors and Fraunhofer IESE
+*
+* SPDX-License-Identifier: MIT
+******************************************************************************/
```
aaronzi, line 24: Fix license header

internal/common/descriptors/postgres_create_batch.go
```diff
@@ -0,0 +1,182 @@
+/*******************************************************************************
+* Copyright (C) 2026 the Eclipse BaSyx Authors and Fraunhofer IESE
+*
+* SPDX-License-Identifier: MIT
+******************************************************************************/
```
aaronzi, line 24: Fix license header

internal/common/descriptors/postgres_bulk_create_batch_test.go
```diff
@@ -0,0 +1,81 @@
+/*******************************************************************************
+* Copyright (C) 2026 the Eclipse BaSyx Authors and Fraunhofer IESE
+*
+* SPDX-License-Identifier: MIT
+******************************************************************************/
```
aaronzi, line 24: Fix license header

internal/common/descriptors/postgres_create_batch.go
```diff
@@ -0,0 +1,182 @@
...
+	"fmt"
+
+	"github.com/FriedJannik/aas-go-sdk/types"
+	"github.com/doug-martin/goqu/v9"
+	"github.com/eclipse-basyx/basyx-go-components/internal/common"
+	"github.com/eclipse-basyx/basyx-go-components/internal/common/model"
+)
+
+// BuildAdministrationShellDescriptorCreateBatch collects all inserts for one
+// Asset Administration Shell descriptor in execution order.
```
aaronzi: Please improve GoDoc

internal/common/postgres_batch.go
aaronzi, line 1: Please improve GoDoc in this file

internal/common/postgres_batch_descriptor.go
aaronzi, line 1: Please improve GoDoc in this file

internal/common/descriptors/postgres_bulk_create_batch.go
```diff
@@ -0,0 +1,539 @@
+/*******************************************************************************
+* Copyright (C) 2026 the Eclipse BaSyx Authors and Fraunhofer IESE
+*
+* SPDX-License-Identifier: MIT
+******************************************************************************/
```
aaronzi, line 24: Fix license header

internal/common/descriptors/postgres_bulk_create_batch.go
```diff
@@ -0,0 +1,539 @@
...
+	ids                       bulkCreateIDs
+	descriptorIndex           int
+	specificAssetIDIndex      int
+	specificSupplementalIndex int
+	submodelSupplementalIndex int
+}
+
+// BuildAdministrationShellDescriptorsCreateBatch collects bulk descriptor
+// inserts while reserving the identifiers needed by dependent rows.
+func BuildAdministrationShellDescriptorsCreateBatch(
```
aaronzi, line 91: Please improve GoDoc here

internal/common/descriptors/postgres_bulk_create_batch.go
aaronzi: Sure, why not

internal/common/descriptors/CompanyDescriptorHandler.go
aaronzi, line 1: Maybe do this kind of refactor in a separate PR

internal/common/descriptors/company_descriptor_error_test.go
```diff
@@ -0,0 +1,26 @@
+/*******************************************************************************
+* Copyright (C) 2026 the Eclipse BaSyx Authors and Fraunhofer IESE
+*
+* SPDX-License-Identifier: MIT
+******************************************************************************/
```
aaronzi, line 24: Fix license header

internal/common/descriptors/AssetAdminShellDescriptorHandler.go
```diff
@@ -377,7 +377,7 @@ func deleteDescriptorRowsBySelectTx(ctx context.Context, tx *sql.Tx, descriptorI
 
 func buildAASDescriptorInsertRecord(
 	ctx context.Context,
-	descriptorID int64,
+	descriptorID any,
```
aaronzi, line 408: Please explain this change

internal/common/descriptors/AssetAdminShellDescriptorHandler.go
```diff
@@ -98,9 +98,9 @@ func InsertAssetAdministrationShellDescriptor(ctx context.Context, db *sql.DB, a
 	return result, tx.Commit()
 }
 
-// canSkipPostInsertReadback returns true for contexts where no post-insert
+// CanSkipPostInsertReadback returns true for contexts where no post-insert
 // descriptor re-read is needed for ABAC enforcement or field filtering.
-func canSkipPostInsertReadback(ctx context.Context) bool {
+func CanSkipPostInsertReadback(ctx context.Context) bool {
```
aaronzi: Since this function is public now, please improve GoDoc

internal/aasxfileserver/persistence/aasx_database.go
aaronzi, line 1: Maybe do this type of refactor in a separate PR?

internal/aasrepository/persistence/aas_database.go
aaronzi, line 1: Refactor for a separate PR?

internal/aasregistry/persistence/PostgreSQLAASRegistryDatabase.go
aaronzi, line 1: Please write better GoDoc for new public functions

aaronzi (changes requested):

go.mod
Martin187187: old posgtres drivider didnt supoort the functionallity i need

internal/submodelrepository/persistence/submodel_write_database.go
Martin187187: this changes was needed. to resolve error code from new db driver

internal/submodelrepository/persistence/submodelElements/submodel_element_database_utils.go
Martin187187: this changes was needed. to resolve error code from new db driver

internal/common/WriteContextReference.go
Martin187187: no its a wrapper so i can reuse this logic, without change the old fucntion

internal/common/error_handler_test.go
Martin187187: the new driver supports the possibility to have have multiple inserts that create indexes that are used int other inserts in the same sql block

internal/common/descriptors/CompanyDescriptorHandler.go
Martin187187: this change was needed to combine error code logic for both both dirvers

internal/common/error_handler_test.go
aaronzi: Shouldn't we then switch to just pgx?

internal/common/descriptors/AssetAdminShellDescriptorHandler.go
Martin187187: return type from new driver is not int64

internal/aasxfileserver/persistence/aasx_database.go
Martin187187: this changes was needed. to resolve error code from new db driver

internal/aasrepository/persistence/aas_database.go
Martin187187: this changes was needed. to resolve error code from new db driver

internal/aasregistry/api/api_asset_administration_shell_registry_api_service.go
Martin187187: yes

internal/common/error_handler_test.go
Martin187187: this would lead to a massive refactor i think in even more files

internal/aasregistry/api/bulk_atomic_operations.go
Martin187187: This is a fair question, but not a bug from this PR. Before the change, ExecuteBulkCreateAtomic delegated to executeAtomicAASDescriptorBulk; with an empty slice, the loop did nothing and returned successfulAtomicResult(0). The new code returns the same result earlier, just without opening a transaction. So the right reply is: “Yes, this was already successful before; this only avoids an empty transaction.” Changing it to bad request would be a behavior/API decision, not a review fix.

internal/aasregistry/api/bulk_atomic_operations_test.go
Martin187187: no, my bad.

aaronzi (approved):

internal/common/descriptors/postgres_bulk_create_batch.go
FriedJannik, line 1: `postgres_bulk_create_batch.go` now combines ID counting, sequence reservation, descriptor flattening, discovery integration behavior, and table insert ordering. Since the cursor methods assume the count/collect logic stays perfectly aligned, a future drift can panic rather than return an error. At minimum, cursor exhaustion validation would make this safer.

internal/aasregistry/api/bulk_atomic_operations.go
FriedJannik, line 1: I think this bulk-create existence check needs another look under ABAC. The old path checked existence through `Get...ByIDInTransaction`, which applies the descriptor read/filter path. The new bulk path queries `aas_descriptor` / `submodel_descriptor` directly and returns `409 Conflict` for any existing ID.

That can leak the existence of descriptors the caller is not allowed to see, and it may change the previous denied/not-found behavior. Can we either apply the same visibility predicate here, or only use this fast existence query when formula/filter enforcement is not active?

internal/aasregistry/persistence/PostgreSQLAASRegistryDatabase.go
```diff
@@ -162,6 +334,125 @@ func (p *PostgreSQLAASRegistryDatabase) InsertAdministrationShellDescriptorInTra
...
+	}
+	batch, err := descriptors.BuildAdministrationShellDescriptorsCreateBatch(ctx, tx, aasDescriptors)
+	if err != nil {
+		return 0, err
+	}
+	if err = common.ExecutePostgreSQLBatchInTransaction(ctx, tx, batch.Statements()); err != nil {
+		return 0, err
+	}
+
+	if descriptors.CanSkipCreateReadback(ctx) {
```
FriedJannik: For AAS bulk create with history enabled, this path can skip readback and write history from the request descriptors. That means the history snapshot may not match the persisted descriptor graph when persistence enriches/normalizes data, for example discovery integration adding `globalAssetId` as a specific asset ID.

Could we force readback when history is active, or build the history snapshot from the same canonical persisted representation?

PR #415, opened 2026-06-26
Removes the need of /tmp dir for /upload, /serialize, /thumbnail and /attachment

FriedJannik:
## Summary

This PR removes the dependency on writable local temp directories for multipart upload handling in the AAS and Submodel repository APIs.

The affected thumbnail and file attachment upload paths now parse multipart requests without `ParseMultipartForm`/temp-file staging and pass upload content to the persistence layer through readers.

## Changes

- Stream multipart file uploads for:
  - AAS thumbnails
  - AAS-scoped submodel file attachments
  - Submodel repository file attachments
- Preserve existing `fileName` form-field behavior regardless of multipart part order.
- Avoid writing upload payloads to `/tmp`.
- Return `413 Payload Too Large` when the configured upload limit is exceeded while the stream is being consumed.
- Keep compatibility for clients that send `fileName` after the file part by buffering that single file in memory under the configured upload limit.
- Add regression tests for:
  - uploads without writable `TMPDIR`
  - oversized upload handling
  - late `fileName` multipart fields

## Notes

When `fileName` appears before the file part, the upload path remains streaming. When `fileName` appears after the file part, the implementation uses bounded in-memory buffering to preserve the previous API contract without falling back to local temp files.

AASX handling is not fully solved by this PR and should be tracked separately as a memory-scalability topic. Ticket #417 describes this in detail. The AASx File Server still uses the temp-file path. Ticket #418 will address this later on.


FriedJannik (commented):
lgmt

PR #431, opened 2026-07-01
Fix panic in TokenizeField on malformed array segments

ffwkfjyvtr-sketch:
Sketch 

`TokenizeField` in `internal/common/builder/field_tokenizer.go` slices `t[open+1 : close]` whenever a segment contains both `[` and `]`. When `]` appears before `[` in a segment (e.g. `a]b[1`), the slice bounds are inverted and the function panics with `slice bounds out of range`.

This PR guards the array-selector parse on `closeIndex > openIndex` and keeps segments without a well-formed selector as `SimpleToken`s. Behavior for all well-formed inputs is unchanged (all existing `TestTokenizeFieldPaths` cases still pass unchanged). The GoDoc example output was also corrected: `submodels[2]` produces a single `ArrayToken`, not a `SimpleToken` plus an `ArrayToken`.

`$field`/fragment strings from the public query API are regex-validated during unmarshaling, so this is primarily defense-in-depth for the other callers of this exported function (e.g. programmatically constructed field paths in the ABAC engine helpers) and for future call sites.

### DIFF
--- a/internal/common/builder/field_tokenizer.go
+++ b/internal/common/builder/field_tokenizer.go
@@ -39,14 +39,18 @@ import (
 // TokenizeField performs the following steps:
 //  1. Removes everything up to and including the first '#'.
 //  2. Splits the remaining expression on '.'.
-//  3. Converts each segment into either a SimpleToken (no brackets)
-//     or an ArrayToken (contains "[index]").
+//  3. Converts each segment into either a SimpleToken (no well-formed
+//     array selector) or an ArrayToken (contains "[index]").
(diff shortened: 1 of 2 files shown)


Martin187187:
Yes, the change is correct as a defensive fix: it prevents TokenizeField from panicking if malformed array syntax like ]0[ reaches it.
That said, ]0[ does not make sense as array syntax in the first place, and valid field identifiers should already be rejected by the grammar/validation layer before reaching the tokenizer.

aaronzi:
Based on the evaluation of @Martin187187, I would close this PR.

PR #432, opened 2026-07-01
Fix nil dereference when updating File element without value

ffwkfjyvtr-sketch:
Sketch

`PostgreSQLFileHandler.Update` (dispatched for `ModelTypeFile` via the handler registry) compared the stored file value against `*file.Value()` unconditionally. `File.value` is optional in the AAS metamodel (`Value()` returns `*string`), so a PUT of a File submodel element without a `value` field dereferences a nil pointer and panics.

In addition, the current value was scanned into a plain `string`. `file_element.value` is NULL for File elements created without a value (`GetInsertQueryPart` inserts `file.Value()` directly), so updating such an element failed with a NULL-conversion error even before the panic could be reached.

The fix scans into `sql.NullString` and compares nil-safely: a nil incoming value is treated as the empty string, matching the semantics already used by `DeleteFileAttachment`, which resets the value to `""`. Behavior for File elements with non-null values is unchanged.

### DIFF
--- a/internal/submodelrepository/integration_tests/submodelrepository_file_put_integration_test.go
+++ b/internal/submodelrepository/integration_tests/submodelrepository_file_put_integration_test.go
@@ -1,0 +1,125 @@
+/*******************************************************************************
+* Copyright (C) 2026 the Eclipse BaSyx Authors and Fraunhofer IESE
+*
+* Permission is hereby granted, free of charge, to any person obtaining
+* a copy of this software and associated documentation files (the
+* "Software"), to deal in the Software without restriction, including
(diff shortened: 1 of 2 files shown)


aaronzi:
Thanks for the fix @ffwkfjyvtr-sketch. Can you please sign the Eclipse Contributor Agreement with the same mail you made your commit with. Otherwise I can't merge your PR.

aaronzi:
@ffwkfjyvtr-sketch Would you be so kind to add a test that can reproduce the problem that existed before.

aaronzi (approved):

PR #487, opened 2026-07-10
make $sme truely recursive

Martin187187:
## Description of Changes

This PR makes `$sme` ABAC fragment filtering row-local and recursive for Submodel Repository read operations.

The main changes are:

- Evaluate structural `$sme` filters against the current SubmodelElement row while retaining submodel-wide existential semantics for regular `$sme` formulas.
- Recursively exclude descendants of filtered parent elements.
- Verify ancestor visibility for exact SME and path requests.
- Apply authorization filters before pagination limits and cursor validation.
- Make exact SME authorization and payload retrieval part of the same database statement, preventing authorization and data retrieval from observing different database states.
- Apply filtering and masking consistently to:
  - direct Submodel responses;
  - Submodel lists and query responses;
  - SubmodelElement responses;
  - `$value`, `$metadata`, `$reference`, and `$path` representations.
- Keep `level=core` requests on the non-recursive path.
- Correct fragment matching across different fragment roots and preserve wildcard-array handling.
- Add a dedicated Submodel Repository security-filter integration test suite to CI.

## Related Issue

No linked issue.

The added security-filter integration environment uses:

```text
ABAC_ENABLED=true
ABAC_MODELPATH=/security_env/access-rules.json
OIDC_TRUSTLISTPATH=/security_env/trustlist.json
```
The complete configuration is located under:

```
internal/submodelrepository/security_filter_tests/
```

The test environment starts the Submodel Repository, PostgreSQL, Keycloak, and the BaSyx Configuration Service through the included Docker Compose configuration.

The security-filter integration tests are executed by the Submodel Repository Security Tests GitHub Actions job.

## AAS Files Used for Testing

The following JSON fixtures are included:

- `internal/submodelrepository/security_filter_tests/bodies/nameplate.json`
  - Tests structural filtering, field masking, exact SME authorization, pagination, and the different SME representations.
- `internal/submodelrepository/security_filter_tests/bodies/deep_path_filter.json`
  - Tests recursive filtering of descendants and visibility through parent/ancestor paths.

Expected filtered responses are stored under:

```
internal/submodelrepository/security_filter_tests/expected/
```

## Additional Information

Structural SME filters for deep/default path responses use recursive SQL queries so that a visible descendant cannot be returned through a filtered parent. Recursive traversal is skipped when no structural SME filter is present, and level=core uses direct row filtering.

A known follow-up concerns the composition of multiple simultaneously applicable ALLOW rules where one rule contains an SME filter and another grants unfiltered access. The currently proven behavior is a false denial: the filtered rule can restrict an otherwise valid unfiltered grant. No authorization overgrant was demonstrated. This should be tracked separately because correcting it requires preserving rule provenance through the authorization-query pipeline.


internal/submodelrepository/persistence/submodelElements/submodel_element_read.go
```diff
@@ -999,6 +987,142 @@ func buildSMEMaskRuntime(ctx context.Context, collector *grammar.ResolvedFieldPa
...
+	}
+	sort.Slice(normalized, func(i int, j int) bool {
+		return normalized[i].from < normalized[j].from
+	})
+
+	fragments := make(map[grammar.FragmentStringPattern]struct{}, len(normalized))
+	for _, filter := range normalized {
+		fragments[filter.to] = struct{}{}
+		delete(clonedQF.Filters, filter.from)
+		if existing, exists := clonedQF.Filters[filter.to]; exists {
```
aaronzi: **Medium:** Structural `$sme` filters are rewritten to `$sme#idShort` here and ANDed with any existing field-level mask. With `$sme: true` and `$sme#idShort: false`, the merged predicate is also used for row visibility, so the whole SME disappears instead of being retained with only `idShort` redacted. Existing query test 058 establishes the field-mask behavior. Please keep structural row predicates separate from field masks and add mixed `$sme`/`$sme#idShort` integration coverage.

internal/submodelrepository/persistence/submodelElements/submodel_element_read.go
```diff
@@ -278,6 +282,10 @@ func GetSubmodelElementPathsPageBySubmodelID(ctx context.Context, db *sql.DB, su
 			return nil, "", common.NewInternalServerError("SMREPO-GETSMEPATHSPAGE-ABACFORMULA " + addFormulaErr.Error())
 		}
 	}
+	query, rowFilterErr := addSMEVisibleTreeQuery(ctx, query, int64(submodelDatabaseID))
```
aaronzi: **Medium:** Cursor validation happens before this new ABAC row/ancestor filter and calls the unfiltered `submodelElementCursorExists`. This creates an existence oracle: with the new fixture, anonymous cursor `QVJlc3RyaWN0ZWQ` (`ARestricted`, hidden) returns the following visible element, while `Qk5vbmV4aXN0ZW50` (`BNonexistent`) returns an empty result. Please validate cursors against the same filtered dataset, or remove the existence check and treat cursors purely as keyset boundaries. A hidden-versus-nonexistent cursor regression test would cover this.

internal/submodelrepository/persistence/submodelElements/submodel_element_read.go
```diff
@@ -999,6 +987,142 @@ func buildSMEMaskRuntime(ctx context.Context, collector *grammar.ResolvedFieldPa
...
+
+	const (
+		visibleCTE   = "visible_sme_ids"
+		rootAlias    = "visible_sme_root"
+		childAlias   = "visible_sme_child"
+		parentAlias  = "visible_sme_parent"
+		visibleIDCol = "id"
+	)
+	dialect := goqu.Dialect("postgres")
+	rootQuery := dialect.
```
aaronzi, line 1045: **Medium:** For exact SME lookups this recursive CTE starts at every root and traverses every visible SME in the submodel before the target-path predicate is applied. The `$path` endpoint invokes the full traversal twice—during authorization and again while fetching paths—turning indexed exact reads into O(all SMEs) work. Please use an upward target-to-ancestor CTE for exact paths and combine authorization with the final path query to avoid the duplicate traversal.

internal/common/model/grammar/logical_expression_to_sql.go
```diff
@@ -136,6 +136,41 @@ func NewResolvedFieldPathCollectorForRoot(root CollectorRoot) (*ResolvedFieldPat
...
 
+// NewResolvedFieldPathCollectorForSMERow creates an SME collector correlated
+// to the current submodel element instead of the containing submodel.
+// Fragment filters use this collector so their conditions are evaluated for
+// each SME row independently, while regular SME formulas retain existential
+// submodel-wide semantics through CollectorRootSME.
+func NewResolvedFieldPathCollectorForSMERow(rootAlias string) (*ResolvedFieldPathCollector, error) {
+	rootAlias = strings.TrimSpace(rootAlias)
+	if rootAlias == "" {
+		return nil, fmt.Errorf("SME row collector root alias must not be empty")
```
aaronzi: **Low:** This new exported helper returns uncoded errors here and for an unsupported base alias, contrary to the repository's mandatory error-prefix convention. Please add stable prefixes such as `GRAMMAR-SMEROWCOLLECTOR-EMPTYALIAS` and `GRAMMAR-SMEROWCOLLECTOR-BADBASE`, and cover the empty-alias path.

internal/submodelrepository/security_filter_tests/docker_compose/docker_compose.yml
```diff
@@ -0,0 +1,99 @@
...
+      KC_HTTP_ENABLED: "true"
+      KC_HTTPS_ENABLED: "false"
+      KC_HEALTH_ENABLED: "true"
+      KC_HOSTNAME_STRICT: "false"
+      KC_HOSTNAME_STRICT_BACKCHANNEL: "false"
+      KC_IMPORT: /opt/keycloak/data/import/
+      KC_HTTP_MANAGEMENT_ENABLED: "true"
+    command: ["start-dev", "--import-realm", "--health-enabled=true"]
+    ports:
+      - "0.0.0.0:${BASYX_IT_KEYCLOAK_PORT:-8080}:8080"
```
aaronzi, line 68: **Low:** This test Keycloak runs in HTTP development mode with `admin/admin` and is published on every host interface, exposing its administration and token endpoints during local test runs. Please bind it to `127.0.0.1:${BASYX_IT_KEYCLOAK_PORT:-8080}:8080`; mounting `/security_env` read-only would also be safer.

aaronzi (commented):
Looks mostly good to me. Please have a look at the following findings below

internal/submodelrepository/persistence/submodelElements/submodel_element_read.go
Martin187187: verified issue with test and fixed

internal/submodelrepository/persistence/submodelElements/submodel_element_read.go
Martin187187: fixed

internal/submodelrepository/persistence/submodelElements/submodel_element_read.go
Martin187187: fixed

aaronzi (commented):
after your latest fixes, I have four more remarks

internal/submodelrepository/persistence/submodelElements/submodel_element_read.go
```diff
@@ -382,6 +373,10 @@ func GetSubmodelElementPathsByPath(ctx context.Context, db *sql.DB, submodelID s
 			return nil, common.NewInternalServerError("SMREPO-GETSMEPATHSBYPATH-ABACFORMULA " + addFormulaErr.Error())
 		}
 	}
+	query, ancestorVisibilityErr := addSMEPathAncestorVisibilityQuery(ctx, query, int64(submodelDatabaseID), idShortPath)
```
aaronzi, line 372: **Medium:** For default/deep `$path` reads, this helper validates only the requested target and its ancestors, while the outer query still returns the target plus every descendant. A reproduced request for a structurally visible collection returned a filtered `HiddenChild` path and a grandchild beneath it. Please retain the upward authorization check but add a target-seeded downward recursive CTE that applies the structural row filter to every descendant and stops recursion through hidden parents. An integration case with a visible parent and hidden child would cover this.

internal/submodelrepository/persistence/submodelElements/submodel_element_read.go
```diff
@@ -991,14 +923,200 @@ func buildSMEMaskRuntime(ctx context.Context, collector *grammar.ResolvedFieldPa
...
+	dialect := goqu.Dialect("postgres")
+	targetQuery := dialect.
+		From(goqu.T("submodel_element").As(targetAlias)).
+		Select(goqu.I(targetAlias+".id"), goqu.I(targetAlias+".parent_sme_id")).
+		Where(
+			goqu.I(targetAlias+".submodel_id").Eq(submodelDatabaseID),
+			goqu.I(targetAlias+".idshort_path").Eq(targetPath),
+		)
+
+	collector, collectorErr := grammar.NewResolvedFieldPathCollectorForSMERow(targetAlias)
```
aaronzi: **Medium:** Regular SME formulas are evaluated with the row-correlated collector here, changing their established submodel-wide existential semantics. In a reproduced policy where one sibling satisfied `$sme#semanticId`, list access succeeded but an exact read of another sibling incorrectly returned 404. This also contradicts `NewResolvedFieldPathCollectorForSMERow`'s GoDoc, which says regular formulas retain `CollectorRootSME` semantics. Please keep this helper limited to structural row filters and apply the regular formula separately with `CollectorRootSME`, with a sibling-satisfies-formula regression test.

internal/common/security/abac_engine.go
```diff
@@ -483,6 +483,9 @@ func (q *QueryFilter) FilterExpressionEntriesFor(key grammar.FragmentStringPatte
 	keyTokens := builder.TokenizeField(string(key))
 
 	for k, expr := range q.Filters {
+		if !fragmentRootsEqual(k, key) {
```
aaronzi, line 486: **Low:** `fragmentRootsEqual` compares the entire prefix before `#`, so `$sme.List[]#value` no longer matches `$sme.List[0]#value`. That contradicts `FilterExpressionEntriesFor`'s documented index-agnostic behavior; a direct regression test returns zero entries instead of one. Please compare only the actual scope token (`$sme`, `$sm`, etc.) before performing the existing tokenized path comparison, and add an SME-path wildcard test.

internal/submodelrepository/persistence/submodelElements/submodel_element_read.go
```diff
@@ -838,33 +810,39 @@ func formatRootCursor(path string, id int64) string {
...
-		if errors.Is(queryErr, sql.ErrNoRows) {
-			return false, nil
-		}
+	rows, queryErr := db.QueryContext(ctx, sqlQuery, args...)
+	if queryErr != nil {
 		return false, common.NewInternalServerError("SMREPO-CHECKSMECURSOR-EXECQ " + queryErr.Error())
 	}
-	return true, nil
+	defer func() { _ = rows.Close() }()
+	return rows.Next(), rows.Err()
```
aaronzi: **Low:** `rows.Err()` can escape from this new helper without the repository's mandatory coded error prefix. Please store the `rows.Next()` result, check and wrap `rows.Err()` with a code such as `SMREPO-CHECKSMECURSOR-ROWSERR`, and add a row-error test.

PR #502, opened 2026-07-19
Add history-independent WORM evidence and deduplicated binary storage

### LINKED ISSUE #416: [Ticket] Store uploaded file and thumbnail binaries in WORM evidence storage when history evidence is enabled
## Description

Currently, when history evidence is enabled, file and thumbnail uploads create history events that are written to WORM evidence storage. However, the WORM artifact only contains the history event metadata/snapshot/diff and hashes. The actual uploaded binary content is still only stored in the regular database storage, for example as PostgreSQL large objects.

This means history evidence can prove that a file-related model state changed, but it cannot restore or verify the uploaded binary itself if the regular file storage is deleted, corrupted, or modified.

## Goal

When history evidence is active, uploaded file attachments and thumbnails should also be stored immutably in WORM/object storage, or at least have their binary content cryptographically anchored there.

## Affected Areas

* Submodel file attachments
* AAS thumbnails
* Possibly AASX/serialization-related attachments, if applicable

## Suggested Implementation

While streaming the upload, calculate a SHA-256 hash of the binary content and write the same stream to immutable object/WORM storage. Store the resulting object reference, object version, digest, size, filename, and content type in the history/evidence metadata.

## Acceptance Criteria

* File attachment uploads store the binary content or immutable binary evidence in WORM storage when history evidence is enabled.
* Thumbnail uploads store the binary content or immutable binary evidence in WORM storage when history evidence is enabled.
* The history event/evidence artifact references the immutable binary object and includes digest, size, content type, and filename where available.
* Upload handling remains streaming-based and does not require writable local temp directories.
* Existing deployments without history evidence enabled keep their current behavior.
Tests cover file attachment and thumbnail uploads with history evidence enabled.

aaronzi:
## Description of Changes

This PR makes WORM evidence independent from PostgreSQL history. Evidence can be enabled with `history.mode: off`, and the same canonical mutation artifact is written whether PostgreSQL history is enabled or not. Existing v1 evidence remains supported by the verifier and recovery tooling.

Internal File attachments and asset-information thumbnails now share content-addressed PostgreSQL Large Object storage based on SHA-256 and byte length. Every upload still receives a fresh owner-scoped `/aasx/files/<token>/<filename>` model path. Identical content reuses the canonical bytes, but the path token and digest are never used as authorization or lookup authorities.

When evidence is enabled, canonical binary content is streamed to immutable object storage and a per-upload reference artifact correlates the owning model path, binary receipt, and mutation event. Full AAS and Submodel replacements preserve an existing managed binary only when the same owner and exact managed path survive the replacement. Removing, externalizing, or changing the path removes the live association and cleans up the PostgreSQL Large Object after its final reference.

AASX serialization keeps managed File and thumbnail paths unchanged and writes the corresponding package parts at those exact URIs. `/aasx/files/...` remains a model and package-part path; no new HTTP route is introduced. The standardized attachment and thumbnail endpoints remain the only public binary access paths.

The final database schema is contained in `database/patches/1_1_8.sql`; `database/base.sql` is unchanged. The patch does not scan, rewrite, deduplicate, or WORM-backfill legacy binaries. Existing paths, table rows, and Large Object OIDs remain untouched until the owning binary is replaced or removed. The upgrade test covers a released pre-PR database and verifies that legacy attachments and thumbnails remain readable.

The follow-up hardening in this branch also:

- uses identifier digests for evidence indexes so identifiers accepted by existing databases do not exceed PostgreSQL B-tree limits;
- acquires evidence and canonical-content locks in stable order for bulk and cascading operations;
- applies resource-level ABAC formulas to attachment uploads and deletes without exposing hidden target types;
- records persisted post-mutation state instead of request objects when persistence normalizes model data;
- documents PostgreSQL and WORM growth, evidence-catalog backup requirements, retention lifecycle behavior, and the accepted global deduplication timing signal.

## Related Issue

Closes #416

The normal integration-test configuration covers canonical PostgreSQL storage and the standardized attachment and thumbnail APIs.

Evidence-only operation can be tested with:

```text
BASYX_HISTORY_MODE=off
BASYX_HISTORY_EVIDENCE_ENABLED=true
BASYX_HISTORY_EVIDENCE_PROVIDER=s3
```

The remaining S3 settings must point to a versioned Object Lock-enabled bucket with a retention mode and positive retention period. The existing history evidence example provides a MinIO-based local setup.

## AAS Files Used for Testing

No additional external AAS files are required. Generated models and the existing repository fixtures cover attachments, thumbnails, history/evidence, AASX round trips, full replacements, legacy upgrades, and shared-content cleanup.

## Additional Information

Important operational points:

- Apply v1.1.8 as a quiesced upgrade and keep a restorable PostgreSQL backup including Large Objects.
- WORM writes remain synchronous and fail closed, so an acknowledged mutation always has its required evidence.
- Evidence-only mode avoids PostgreSQL snapshot and diff payload growth, but receipt metadata still grows with mutations. WORM stores one mutation artifact per mutation and one reference artifact per internal upload; unique binary bytes are deduplicated.
- Object Lock retention expiry does not delete objects by itself. Eventual deletion requires an operator-managed lifecycle policy, and evidence receipt rows are not pruned automatically.
- The built-in mutation verifier uses the PostgreSQL evidence catalog to locate committed immutable object versions, so the catalog must be part of backup and recovery procedures.
- Global deduplication can expose a noisy timing signal to an already-authorized uploader. Deployments that require cross-tenant timing isolation should use separate databases or service instances.
- Future eventing can reuse the same mutation boundaries, but reliable delivery should use a transactional outbox rather than publishing from the open model transaction.

Validation performed:

- `golangci-lint run`
- `go test ./...`
- `go test -v ./internal/aasenvironment/migration_integration_tests`
- `go test -v ./internal/aasenvironment/integration_tests`
- `go test -v ./internal/aasrepository/integration_tests`
- `go test -v ./internal/submodelrepository/security_tests`
- cache-cleared `go test -v ./internal/submodelrepository/integration_tests`


docu/security/NIS2_HISTORY_EVIDENCE.md
```diff
@@ -20,7 +21,11 @@ BaSyx provides technical controls that can support NIS2-aligned integrity, audit
...
 - Generate, rotate, back up, and distribute signing and verification keys through controlled trust processes.
 - Monitor verifier findings and alert on critical drift or signature failures.
+- Retain verified mutation sequence/hash heads outside the BaSyx PostgreSQL database and advance them only after successful verification. Protect the first trust-on-first-use baseline through an operator-controlled process.
+- Back up and restore the PostgreSQL evidence catalog with the database. The built-in mutation verifier uses it to locate committed WORM objects and their immutable versions.
+- Configure object-store lifecycle deletion separately when required. Object Lock retention expiry permits deletion but does not perform it; keep lifecycle periods aligned with legal and operational verification windows.
 - Manage vulnerability handling, patching, and dependency updates.
+- Use a controlled, quiesced v1.1.8 upgrade with a verified PostgreSQL backup that includes Large Objects. Do not mix v1.1.7 and v1.1.8 database writers or attempt a binary-only rollback.
```
FriedJannik:
```suggestion
- Use a controlled, quiesced v1.1.8 upgrade with a verified PostgreSQL backup that includes Large Objects.
```

Components check their DB Version and fail if DB Version > expected

internal/aasenvironment/serialization_api_service.go
```diff
@@ -622,14 +622,18 @@ func (s *SerializationAPIService) resolveSerializationSupplementaryParts(
 			continue
 		}
 
-		resolvedReference := ResolveAASXReferenceAgainstSpec(fileLocation.FileValue, specURI)
+		managedReference := strings.HasPrefix(fileLocation.FileValue, serializationAASXSupplementaryRoot+"/")
```
FriedJannik:
```suggestion
		isManagedReference := strings.HasPrefix(fileLocation.FileValue, serializationAASXSupplementaryRoot+"/")
```

internal/aasrepository/persistence/aas_database.go
```diff
@@ -1000,13 +1060,81 @@ func (s *AssetAdministrationShellDatabase) putAssetAdministrationShellByIDInTran
...
 	return isUpdate, nil
 }
 
+type managedThumbnailForReplacement struct {
+	managedPath  string
+	contentType  sql.NullString
+	contentID    int64
+	pathToken    string
+	safeFileName string
+}
```
FriedJannik: Please extract this to
common>model>transactionObjects.go

internal/aasrepository/persistence/thumbnail_file_handler.go
```diff
@@ -437,3 +441,254 @@ func (h *PostgreSQLThumbnailFileHandler) deleteThumbnailByAASIDInTransaction(tx
 
 	return nil
 }
+
+type managedThumbnailMetadata struct {
+	aasDBID             int64
+	existingContentType sql.NullString
+	existingFileName    sql.NullString
+	path                sql.NullString
+}
```
FriedJannik: Please extract to
common>model>transactionObjects

internal/common/history/mutation_evidence.go
FriedJannik, line 1: Please extract Database DTO to transferObjects.go

internal/submodelrepository/persistence/submodel_write_database.go
```diff
@@ -575,10 +612,147 @@ func (s *SubmodelDatabase) replaceSubmodelInTransaction(tx *sql.Tx, submodelID s
...
 	return true, nil
 }
 
+type managedFileReferenceForReplacement struct {
+	idShortPath  string
+	managedPath  string
+	contentID    int64
+	pathToken    string
+	safeFileName string
+}
```
FriedJannik: move to transferObjects.go

FriedJannik (changes requested):

FriedJannik (approved):
Approved. Thank you for implementing :)

PR #508, opened 2026-07-23
fix: delegated operation outputArguments returned as [{}]

Encotric:
## Problem

Delegated operation responses could serialize `outputArguments` as empty objects because `types.OperationVariable` cannot be marshalled directly with `encoding/json`. The asynchronous invocation implementation also followed the older handle-in-body flow instead of the AAS V3.2 `Location`-based protocol.

## Changes

- Convert delegated operation variables through the AAS JSONization layer so input, output, and inoutput arguments retain their values.
- Implement the AAS V3.2 asynchronous invocation lifecycle:
  - `POST /invoke-async` returns `202 Accepted` and a status `Location`.
  - running status requests return `200 OK`.
  - completed status requests return `302 Found` and a result `Location`.
  - result requests return the normalized `OperationResult`.
- Use the official V3.2 `OperationRequestAsync` schema and require `clientTimeoutDuration` on `/invoke-async`.
- Keep `POST /invoke?async=true` as a compatibility alias with its default asynchronous timeout.
- Generate correct external `Location` URLs for both Submodel Repository and AAS Repository routes.
- Preserve and normalize optional execution state, success, messages, output arguments, and inoutput arguments from delegated results.
- Return coded gateway errors for malformed delegated responses.
- Bound delegated response bodies to 1 MiB.
- Reject invalid, overflowing, non-positive, and excessive delegation timeouts.
- Bound concurrent asynchronous delegations, retain handles through the complete execution window, renew terminal-result retention, and cancel work during service shutdown.
- Use request-scoped structured logging and preserve the security context for asynchronous work.
- Expose the `Location` response header through CORS so browser clients can use the V3.2 flow.
- Harden the delegated operations example and add a CI protocol smoke test for synchronous invocation and both asynchronous repository route namespaces.

## Impact

Delegated operation results retain their values, asynchronous invocation conforms to AAS V3.2, and browser clients can follow the complete status/result flow. Existing synchronous clients and the legacy async query parameter remain supported.

Async and bulk handles remain process-local. Durable shared persistence and restart/multi-replica behavior are intentionally tracked separately in #525.

## Validation

- All Go unit and integration packages
- Full Submodel Repository integration suite from a clean test cache
- `go vet ./...`
- `golangci-lint`
- Delegated service and shell checks
- Production UI, AAS Environment, and configuration image builds
- Live Docker Compose protocol smoke through the Submodel Repository and AAS Repository namespaces using locally built images

## Related

- Web UI implementation: eclipse-basyx/basyx-aas-web-ui#1424
- eclipse-basyx/basyx-aas-web-ui#1414
- Follow-up for durable async and bulk handles: #525


aaronzi:
Please sign the Eclipse Contributor Agreement with the same mail you used to sign your commits.

Encotric:
Should be done.

aaronzi (approved):

PR #522, opened 2026-07-27
add row wise filter fragment logic for lists in QL

### LINKED ISSUE #376: [Ticket] Implement AAS 3.2 QL row-wise and entire-array filtering for array fragments
## User Story

> **As a** BaSyx user/developer,  
> **I want** QL filters to distinguish between row-wise array filtering and entire-array filtering,  
> **so that** I can filter array or collection-type objects according to the upcoming AAS 3.2 QL filter semantics.

---

## Rules

- **What:**
  - Support fragment field identifiers for array or collection-type objects with and without trailing `[]`.
  - A `$fragment` ending with `[]` applies the condition **row-wise** to each array element.
  - A `$fragment` without trailing `[]` applies the condition to the **entire array object**.
  - Every fragment field identifier ending in `[]` must also be valid without the trailing `[]`.
  - Invalid fragment/field combinations must be rejected with a clear validation error.

- **When:**
  - This logic applies whenever a QL filter uses a `$fragment` that references an array or collection-type object.

- **How:**
  - Extend QL parsing, validation, and evaluation to detect whether the `$fragment` ends with `[]`.
  - Resolve `$field` identifiers inside the condition according to the selected mode:
    - row-wise mode for `$fragment: "...[]"`
    - entire-array mode for `$fragment: "..."`
  - Keep existing QL filter behavior backward compatible.

---

### Row-wise filter

The condition is evaluated for each element of `a`.

```json
{
  "$condition": { "$boolean": true },
  "$filters": [
    {
      "$fragment": "$sme.a[]",
      "$condition": {
        "$eq": [
          { "$field": "$sme.a[]#value" },
          { "$strVal": "val" }
        ]
      }
    }
  ]
}
```
Open to discussion:

- Only support SubmodelArrayList with this logic? Because:
  - SubmodelElementList: Kinder haben Pfade wie List[0], List[1].
  - SubmodelElementCollection: Kinder haben Pfade wie Collection.ChildA, Collection.ChildB.

### LINKED ISSUE #496: [BUG] Filtering `$aas#submodels[]` causes a 500 for an otherwise permitted AAS
**Describe the bug**

Filtering individual references in `AssetAdministrationShell.submodels` with an ABAC `FILTERLIST` makes reading an otherwise permitted AAS fail with HTTP 500.

The AAS access rule itself is allowed and the AAS descriptor can still be listed/read. The failure happens while the repository reads the AAS's submodel-reference payloads. Removing only the `$aas#submodels[]` filter makes the same request return 200 again.

**To Reproduce**

This does not need an IdP, browser, or UI. I reproduced it as the anonymous caller in an AAS Environment.

1. Start an AAS Environment with ABAC enabled and an AAS that has at least two submodel references.
2. Allow `READ` on the parent AAS, then add a filter which should retain only a selected reference:

```json
{
  "USEACL": "anonymous_read",
  "USEOBJECTS": ["all_aas_identifiables"],
  "USEFORMULA": "is_public_aas",
  "FILTERLIST": [
    {
      "FRAGMENT": "$aas#submodels[]",
      "USEFORMULA": "is_public_submodel_reference",
      "MATCH": true
    }
  ]
}
```

The filter formula used in my test was:

```json
{
  "name": "is_public_submodel_reference",
  "formula": {
    "$eq": [
      { "$field": "$aas#submodels[].keys[].value" },
      { "$strVal": "urn:example:submodel:visible" }
    ]
  }
}
```

3. Request the permitted AAS, for example:

```sh
curl -i http://localhost:8081/shells/<base64url-aas-id>
```

4. The response is HTTP 500 instead of the filtered AAS.

I also reproduced this with `IESEDriveMotorDM3000.aasx` from the repository's secured example, filtering its five submodel references down to Nameplate and TechnicalData.

**Expected behavior**

The request should return HTTP 200. The parent AAS should remain readable and its `submodels` array should contain only the references that pass the filter.

**BaSyx (please complete the following information):**
- BaSyx Version: `eclipsebasyx/aasenvironment-go:SNAPSHOT`, image revision `e46abe551fce7f6413e038df74b2409c90303fff` (`SNAPSHOT-20260713-191705-e46abe5`)
- BaSyx SDK: Go / AAS Environment Service
- AAS Version: AAS V3

**Screenshots**

Not applicable; this reproduces with `curl`.

**Used Asset Administration files (.aasx/.json/.xml)**

`IESEDriveMotorDM3000.aasx` from `examples/BaSyxSecuredExample/aas/` was used for the concrete reproduction. Any AAS with two or more submodel references is sufficient.

**BaSyx configuration**

ABAC is enabled with the policy imported from `access-rules.json`. The parent AAS is granted anonymous `READ`; the Submodel resources have separate matching rules. The failure is specifically triggered by the `FILTERLIST` fragment `$aas#submodels[]` (adding the corresponding nested-keys filter is not required to reproduce it).

**System (please complete the following information):**
 - OS: macOS
 - CPU architecture: arm64
 - Browser: Brave (not involved in the reproduction)
 - Version: current local snapshot image

**Additional context**

The response body is:

```text
AASREPO-READSMREFBATCH-EXECSQL ERROR: missing FROM-clause entry for table "aas" (SQLSTATE 42P01)
```

The service code applies the filters in `readSubmodelReferencePayloadsByAASDBIDs` (`internal/aasrepository/persistence/aas_database.go`) before executing the batched submodel-reference query. It looks as though the generated condition still refers to the AAS root alias even though that query does not include it.

As a temporary workaround, I removed the AAS-reference `FILTERLIST`. This restores the endpoint, but it leaves hidden references visible in the raw AAS payload, so it is not suitable as the long-term security behaviour.

Martin187187:
## Summary

Implements AAS 3.2 row-wise filtering semantics for array fragments and fixes the `$aas#submodels[]` filtering regression.

Fragments ending exactly in `[]` are now evaluated for each candidate array row. Fragments without a trailing `[]` retain parent/whole-object filtering semantics. The exact `$sme` fragment is evaluated for each SubmodelElement, including recursively loaded nested elements.

## Changes

- Add fragment-aware SQL generation:
  - evaluate candidate-local fields against the current row;
  - use correlated `EXISTS` expressions for fields outside the candidate scope;
  - keep predicates inside `$match` bound to the same array row.
- Fix the missing AAS correlation that caused `$aas#submodels[]` filters to fail with HTTP 500.
- Apply row-wise filtering consistently across:
  - AAS submodel references;
  - AAS and Submodel descriptors;
  - descriptor endpoints;
  - specific asset IDs and external-subject references;
  - semantic and supplemental semantic references;
  - Submodels and nested SubmodelElements.
- Preserve recursive `$sme` filtering without allowing sibling or descendant matches to expose the wrong element.
- Retain the obsolete `MATCH` field for legacy policy compatibility:
  - `MATCH: true` is accepted for fragments ending in `[]`;
  - `MATCH: false` is accepted for non-array fragments;
  - mismatched combinations are rejected;
  - `MATCH` no longer changes filtering behavior.
- Update the OpenAPI schema, documentation, examples, and security configurations for automatic row-wise filtering.
- Add extensive unit and integration coverage for row scoping, SQL correlation, reference reconstruction, nested SMEs, and legacy compatibility.

## Testing

Added and updated automated coverage for:

- production repository filtering across AAS, descriptor, Submodel, and SME resources;
- row-local versus parent-scope SQL generation;
- `$match` conditions targeting the same array row;
- recursive and exact `$sme` filtering;
- semantic and supplemental semantic reference filtering;
- legacy `MATCH` validation.

## Related Issues

Closes #496  
Closes #376


Martin187187:
todos:

- [x] benchmarks
- [x] decide if legacy MATCH will be rejected or not

Martin187187:
> todos:
> 
> * [x]  benchmarks
> * [x]  decide if legacy MATCH will be rejected or not

- scalability tests passed
- MATCH behaviour works like this now:

`MATCH` is obsolete and has no effect on filtering. Legacy policies may still
include it when its value agrees with the fragment: `MATCH: true` requires a
fragment ending exactly in `[]`, while `MATCH: false` requires a fragment that
does not end in `[]`.

PR #557, opened 2026-08-03
Support explicit matching for fragment filters

### LINKED ISSUE #496: [BUG] Filtering `$aas#submodels[]` causes a 500 for an otherwise permitted AAS

Martin187187:
closes #496

## Overview

This PR makes fragment filtering predictable across query endpoints and ABAC
rules. Both now use the same QL behavior: filters define which fragment and
condition to apply, while `$match` in queries or `MATCH` in rules explicitly
chooses row-local evaluation.

It also fixes nested Submodel Element List correlation and ensures that match
settings remain attached to the correct condition when filters or ABAC rules are
combined.

## What changed

- Added optional `"$match"` support to query filters and the AAS 3.2 OpenAPI
  schema. ABAC rules continue to use the equivalent uppercase `MATCH` property.
- Reworked fragment filters so every condition keeps its own match setting
  through `AND` and `OR` combinations.
- Applied the behavior consistently to descriptor references, endpoints,
  specific asset IDs, semantic references, and submodel elements.
- Fixed nested SME `[]` paths with anchored index matching, preventing sibling
  paths or list entries from satisfying the current row's condition.
- Added unit and integration coverage, updated the query/security docs, and
  added a practical input/expected-result example guide.

## Impact

- With `$match: true` or `MATCH: true`, only fragment rows satisfying that
  condition are returned.
- With match omitted or `false`, the condition is evaluated at parent scope. A
  matching sibling can keep the complete fragment in the response.
- Request queries with an array-ended `$fragment` previously enabled row-local
  matching implicitly. Clients that rely on the narrowed array response must
  now add `"$match": true`. Existing request bodies remain valid, but returned
  fragment contents can be broader when it is omitted.
- ABAC filters remain mandatory. Query filters are ANDed with them, so clients
  can narrow policy-visible data but cannot widen access.
- No endpoint, database schema, persisted-data, or configuration migration is
  required. The affected reads are in the AAS Repository, AAS Registry,
  Submodel Registry, and Submodel Repository.

## Why this is useful

- Evaluation scope is explicit instead of being inferred from a trailing `[]`.
- Match settings no longer leak between combined policy conditions.
- Nested list filtering is correctly bound to the current SME path and index.
- Query endpoints and ABAC policies now behave consistently and are easier to
  reason about and document.


internal/common/swagger_part2_schemas/V3.2.0/openapi.yaml
```diff
@@ -892,6 +892,12 @@ components:
       properties:
         $fragment:
           $ref: '#/components/schemas/FragmentFieldIdentifier'
+        $match:
+          type: boolean
+          default: false
+          description: >-
+            Evaluate the condition against the current fragment row. Omitted or
+            false uses parent-level existential evaluation.
```
aaronzi, line 900: Is this in the official v3.2 part 2 schema?

Martin187187: no, this is something we have to bring into the specification. this is a missing component that was overlooked and therefore has to be introduced via a bug into the spec.

aaronzi (approved):

PR #591, opened 2026-08-12
increase performance repositories put

### LINKED ISSUE #565: [Ticket] Skip unchanged Submodel PUT work and unnecessary registry synchronization
## User Story

> **As a** BaSyx user updating Submodels,  
> **I want** an unchanged full PUT to avoid rebuilding the complete Submodel graph,  
> **so that** retries and idempotent synchronization do not create unnecessary database load.

## Problem

`PutSubmodelByID` currently replaces an existing Submodel by:

1. loading managed file references;
2. cleaning up large objects;
3. deleting the complete Submodel and its SME graph;
4. recreating the Submodel, all SME rows, type rows, payload rows, semantic references and qualifiers;
5. restoring managed file references.

When Submodel Registry integration is enabled, the AAS Environment then rebuilds the Submodel Descriptor and synchronizes referencing AAS Descriptors in the same transaction.

The current capacity update profile intentionally sends the same representation that is already stored. Despite being a semantic no-op, it measured about 5.7 RPS at concurrency 5 and 11.5 RPS at concurrency 10 with roughly 850 ms median latency on the million-record HARTING-derived dataset.

## Scope

- Detect whether the incoming full Submodel representation is semantically equal to the stored representation before deleting data.
- Keep all authorization and visibility checks before returning the successful PUT response.
- If the Submodel is unchanged, skip the relational delete/recreate path and managed file cleanup/restore work.
- In the AAS Environment, skip Submodel Descriptor upsert and referencing AAS Descriptor synchronization when their derived content is unchanged.
- Use one deterministic definition of equality for every valid Submodel and SME type, independent of JSON map order.
- Keep create-vs-update response behavior unchanged.
- Define and test the behavior when history evidence is enabled.
- Add separate benchmark profiles for unchanged PUT and a PUT that changes an actual value.

## Relevant Entry Points

- `internal/submodelrepository/persistence/submodel_write_database.go`
- `internal/aasenvironment/custom_submodel_repository_service.go`
- `internal/aasenvironment/registry_sync.go`
- `internal/common/history/`
- Submodel Repository integration and security tests
- Kubernetes scalability harness from #534

## Acceptance Criteria

- [ ] A semantically unchanged Submodel PUT does not delete or insert Submodel/SME persistence rows.
- [ ] Unchanged PUT does not clean up and restore managed file attachments.
- [ ] Registry integration does not rewrite an unchanged Submodel Descriptor or unchanged embedded descriptor state.
- [ ] ABAC behavior does not reveal whether inaccessible content is equal to the request.
- [ ] Equality works for arbitrary nesting, every supported SME type, optional fields, language strings, references, blobs, files and empty collections/lists.
- [ ] JSON object order and Go map iteration order do not produce false changes.
- [ ] Create, changed update, ID mismatch and denied update behavior stays unchanged.
- [ ] History-enabled and history-disabled behavior is explicitly tested.
- [ ] Benchmarks report unchanged PUT separately from changed PUT and show a clear reduction in database statements and latency.

## Risks and Assumptions

- Loading and comparing a complete large Submodel can still be expensive. The selected change detector must be cheaper than full reconstruction and must not rely on an unstable serialization.
- A hash may be stored or calculated, but hash collision handling and schema migration/backfill must be considered if it becomes correctness-relevant.
- Server-managed administration timestamps must not turn every retry into a false change.
- This ticket handles complete no-ops. Partial structural changes are handled separately.

## Dependencies and Blockers

- Tracked by #564.
- No implementation blocker.
- #566 should reuse the equality/change-detection rules established here.

### LINKED ISSUE #566: [Ticket] Replace only changed Submodel element subtrees on full PUT
## User Story

> **As a** BaSyx user updating large Submodels,  
> **I want** a full PUT to rewrite only the parts that changed,  
> **so that** update cost depends on the actual change instead of the total Submodel size.

## Problem

After unchanged PUTs are handled by #565, every real change still causes the complete Submodel and all Submodel Elements to be deleted and recreated.

That behavior is correct as replacement semantics, but it creates heavy write amplification for realistic Submodels. A change to one Property can rewrite hundreds or thousands of unrelated SME, payload, semantic-reference, qualifier and type rows. Cascading deletes also create index and WAL work and require managed attachments to be cleaned up and restored.

The relational schema already supports targeted SME PUT/PATCH operations. Full Submodel PUT should reuse that capability at a safe granularity instead of rebuilding the complete graph.

## Scope

- Keep the existing Submodel database row and update Submodel metadata in place.
- Compare the incoming and stored top-level SME roots using the deterministic change rules from #565.
- Preserve unchanged top-level subtrees.
- Delete and recreate only changed, removed or newly added top-level subtrees.
- Preserve PUT replacement semantics for omitted fields and removed elements.
- Keep managed files and PostgreSQL large objects attached to unchanged subtrees.
- Clean up managed content only when its owning File/Blob element is changed or removed.
- Keep one transaction and one stable pre-mutation snapshot for authorization and history.
- Support arbitrary valid Submodel structures, nesting depth and SME combinations.
- Do not implement a fine-grained row-by-row diff inside a changed root in the first iteration unless measurements show that top-level subtree replacement is insufficient.

## Relevant Entry Points

- `internal/submodelrepository/persistence/submodel_write_database.go`
- `internal/submodelrepository/persistence/submodel_element_write_database.go`
- `internal/submodelrepository/persistence/submodelElements/`
- `internal/submodelrepository/persistence/submodel_history_snapshot.go`
- managed file and large-object persistence
- Submodel Repository integration, security, extent and file tests

## Acceptance Criteria

- [ ] Changing one top-level SME does not rewrite unrelated top-level SME subtrees.
- [ ] Submodel metadata is updated without replacing the Submodel database row.
- [ ] Added, removed, renamed and type-changed roots produce the same API result as full replacement.
- [ ] Nested collections, lists, entities, operations and annotated relationships are handled correctly.
- [ ] Unchanged managed files keep their references and binary content.
- [ ] Removed or changed managed files do not leak PostgreSQL large objects.
- [ ] ABAC checks, formula and fragment behavior, history snapshots and timestamps stay correct.
- [ ] Existing integration, security, AAS Environment and Catena-X example tests continue to pass.
- [ ] A changed-value benchmark shows lower latency, fewer written rows and less WAL than complete replacement.
- [ ] Read endpoint performance does not regress.

## Risks and Assumptions

- Top-level `idShort` is the initial matching key. Renames and model-type changes must be treated as remove plus add.
- List order and nested paths can change when a parent changes. Replacing the affected root is safer than attempting a global row-level merge.
- The optimization must not keep fields omitted by PUT.
- Fine-grained diffing below top-level roots remains a later option if very large single-root Submodels are still too expensive.

## Dependencies and Blockers

- Tracked by #564.
- Depends on #565 for deterministic equality/change detection and separate unchanged-vs-changed benchmark coverage.

### LINKED ISSUE #570: [Ticket] Avoid delete-and-recreate updates in the AAS Repository
## User Story

> **As a** BaSyx user updating Asset Administration Shells,  
> **I want** the AAS Repository to update an existing shell in place,  
> **so that** small changes do not rebuild the complete shell graph and synchronized descriptor.

## Problem

`PutAssetAdministrationShellByID` currently loads managed thumbnail state, deletes the complete AAS row with cascading child data, recreates the AAS graph and restores the thumbnail reference.

In the AAS Environment, the same transaction also upserts the AAS Descriptor whenever registry integration is active. This repeats work even when descriptor-relevant fields are unchanged.

AAS graphs are usually smaller than Submodels, but the current path still generates avoidable delete, insert, index, WAL and synchronization work for every update.

## Scope

- Preserve the existing AAS database row on update.
- Update core AAS and Asset Information fields in place.
- Replace or diff bounded child collections such as Specific Asset IDs and Submodel references without deleting the AAS root.
- Preserve managed thumbnail ownership and binary content when the thumbnail did not change.
- Skip the persistence write when the complete AAS is unchanged, if that is cheaper than the current replacement.
- In the AAS Environment, update the AAS Descriptor only when descriptor-relevant AAS/Asset Information fields changed.
- Preserve full PUT replacement semantics, ABAC behavior, history and transaction atomicity.

## Relevant Entry Points

- `internal/aasrepository/persistence/aas_database.go`
- `internal/aasrepository/persistence/thumbnail_file_handler.go`
- `internal/aasenvironment/custom_aas_repository_service.go`
- `internal/common/descriptors/`
- AAS Repository integration, security, history and thumbnail tests

## Acceptance Criteria

- [ ] Updating an existing AAS does not delete the AAS root row.
- [ ] Omitted optional fields and removed child values are still removed as required by PUT.
- [ ] Unchanged Submodel references and Specific Asset IDs are not needlessly rewritten.
- [ ] An unchanged managed thumbnail keeps its binary object and reference.
- [ ] Changed or removed thumbnails are cleaned up without leaking large objects.
- [ ] AAS Environment registry integration skips an unchanged derived descriptor.
- [ ] ABAC, history, timestamps, create-vs-update responses and conflict behavior remain unchanged.
- [ ] Existing AAS Repository and AAS Environment integration tests pass.
- [ ] Before/after evidence reports statements, WAL, latency and throughput for unchanged and changed AAS PUTs.

## Risks and Assumptions

- Asset Information and thumbnail persistence share identifiers with the AAS root. In-place updates must preserve those relationships.
- Submodel reference ordering must remain API-compatible.
- Descriptor synchronization is a correctness requirement when descriptor-relevant fields change; it must not be made eventually consistent by this ticket.

## Dependencies and Blockers

- Tracked by #564.
- Independent of the Submodel work.
- #565 can provide reusable ideas for semantic no-op detection, but is not a blocker.

### LINKED ISSUE #590: [BUG] Concurrent identical AAS PUT requests can return HTTP 409
## Describe the bug

Concurrent, identical requests to
`PUT /shells/{aasIdentifier}` can intermittently return HTTP `409 Conflict`
when they target the same existing Asset Administration Shell.

In the observed run, one PUT for
`urn:iese:basyx:harting:v1:aas:0008884` completed with HTTP `204`, while an
overlapping PUT to the same URI returned HTTP `409`. PostgreSQL reported that
the failing request attempted to insert the already-existing AAS identifier:

```text
ERROR: duplicate key value violates unique constraint "aas_aas_id_key"
DETAIL: Key (aas_id)=(urn:iese:basyx:harting:v1:aas:0008884) already exists.
STATEMENT: INSERT INTO "aas" ("aas_id", "category", "id_short")
           VALUES ('urn:iese:basyx:harting:v1:aas:0008884', NULL,
                   'HARTING_AAS_02440000001') RETURNING "id"
```

The two relevant AAS Repository access-log entries are nearly simultaneous:

```text
2026-08-12T09:22:37.744Z PUT .../shells/<base64url-id> 204 duration_ms=138.892
2026-08-12T09:22:37.746Z PUT .../shells/<base64url-id> 409 duration_ms=5.862
```

This suggests a race in the replacement/update persistence path: overlapping
updates can both reach an insert operation, and one loses on the unique AAS ID
constraint. This explanation is an inference from the service and PostgreSQL
logs; it has not yet been confirmed by tracing the server code.

## To reproduce

The issue was observed with the benchmark tooling and Compose configuration in
this repository.

1. Start the local BaSyx Compose stack using the `SNAPSHOT` images:

   ```powershell
   python benchmarkctl.py setup
   ```

2. Configure and seed 10,000 deterministic HARTING-derived resources:

   ```powershell
   python benchmarkctl.py config set harting.cardinality 10000
   python benchmarkctl.py config set harting.seedStateDir results/harting-local-seed-10k
   python benchmarkctl.py seed
   ```

3. Configure a hot-key AAS update workload. The small request pool is
   intentional because it causes workers to revisit the same identifiers:

   ```powershell
   python benchmarkctl.py config set benchmark.profiles update_shell_by_id
   python benchmarkctl.py config set benchmark.requestPoolSize 10
   python benchmarkctl.py config set benchmark.concurrency 5
   python benchmarkctl.py config set benchmark.warmup 10s
   python benchmarkctl.py config set benchmark.duration 30s
   ```

4. Run the benchmark:

   ```powershell
   python benchmarkctl.py run
   ```

5. Inspect the measured result and the AAS Repository/PostgreSQL logs. The
   observed run produced:

   ```text
   requests:       2155
   succeeded:      2154
   failed:         1
   status counts:  204=2154, 409=1
   failure rate:   0.046404%
   throughput:     71.75 requests/s
   concurrency:    5
   request pool:   10
   protocol:       HTTP/1.1
   ```

The warm-up immediately before the measured phase completed 1,127 requests
without a failure. The conflict is timing-dependent, so a run may need to be
repeated. A direct test that sends several concurrent identical PUT requests to
one seeded AAS URI should increase the probability of reproduction.

## Expected behavior

Concurrent identical PUT requests without conditional headers should not fail
because an internal replacement operation races on the resource's own unique
identifier. The server should perform the replacement atomically or serialize
updates to the same AAS. For identical payloads, each request should return a
successful response such as HTTP `204`, with the requested representation as
the final resource state.

If concurrent updates are intentionally rejected, the behavior should be based
on an explicit concurrency-control contract, such as a precondition or version,
rather than an internal duplicate-key insertion error.

## BaSyx

- **BaSyx version:** `SNAPSHOT-20260804-171146-903964b`
- **Source revision:** `903964b34b9f0144b5efadd812e8b1805a262fa3`
- **Container image:** `docker.io/eclipsebasyx/aasrepository-go:SNAPSHOT`
- **Container image ID:**
  `d1ce2769f273e2a02c0367b0a96f563801b6c966cd9deee7d5f83a49ccd71a1a`
- **Image digest:**
  `sha256:41fd1a475039fc577a2eaf9340d5da59381045ce3eff3edace4eea182a2bad3f`
- **BaSyx SDK/components:** BaSyx Go Components
- **AAS version:** DotAAS V3

## Screenshots

No screenshot is required. Attach these text artifacts from the failed run:

- `results/local-put-updates-20260812-092134/config.json`
- `results/local-put-updates-20260812-092134/suite.json`
- `results/local-put-updates-20260812-092134/runs/update_shell_by_id/warmup.json`
- `results/local-put-updates-20260812-092134/runs/update_shell_by_id/measured.json`
- the AAS Repository and PostgreSQL log excerpts shown above

## Used Asset Administration files (.aasx/.json/.xml)

The target was generated deterministically from the prepared HARTING catalog;
no AASX attachment binaries were uploaded during seeding.

- **Generated AAS ID:** `urn:iese:basyx:harting:v1:aas:0008884`
- **Generated AAS idShort:** `HARTING_AAS_02440000001`
- **Reduced source catalog SHA-256:**
  `e86f4cfb953a9eecab0f4749cbd04f59b11f292b23bfa9b44fd17d6b12ff951e`
- **Reduced source catalog contents:** 10 shells, 58 Submodels, and 265
  Concept Descriptions, expanded deterministically to cardinality 10,000

The reduced prepared catalog can be attached separately if required. It is
approximately 52 KiB compressed.

## BaSyx configuration

- PostgreSQL 18 backend
- Nginx `1.27-alpine` reverse proxy at `http://localhost:8082`
- AAS Repository context path: `/aas-repository`
- `GENERAL_AASREGISTRYINTEGRATION=true`
- `GENERAL_DISCOVERYINTEGRATION=false`
- `SERVER_STRICTVERIFICATION=permissive`
- `POSTGRES_MAXOPENCONNECTIONS=500`
- `POSTGRES_MAXIDLECONNECTIONS=500`
- History disabled
- Eventing disabled
- ABAC disabled
- All BaSyx Go services use `SNAPSHOT` with Compose `pull_policy: always`

The complete reproducible configuration is in:

- `local-basyx-podman/docker-compose.yml`
- `local-basyx-podman/nginx/default.conf`
- `benchmarkctl.example.json`

Credentials should be removed or replaced before attaching configuration to a
public issue.

## System

- **Host OS:** Microsoft Windows NT `10.0.26200.0`
- **Host architecture:** x86-64
- **Container VM:** WSL 2, kernel `6.6.87.2-microsoft-standard-WSL2`
- **Podman VM distribution:** Fedora 42
- **Podman client/server:** `5.6.2` / `5.6.2`
- **Container architecture:** `linux/amd64`
- **Available container VM resources:** 12 CPUs, approximately 7.6 GiB RAM
- **Browser:** Not applicable; requests were sent by the Go benchmark client
  through Nginx using HTTP/1.1

## Additional context

The benchmark's failure threshold was `0.0`, so the single HTTP `409` correctly
stopped the suite before the Submodel and SubmodelElement PUT profiles ran.

Increasing the request pool from 10 to 1,000 avoids frequent same-key overlap
and is appropriate for general endpoint-capacity measurements, but it only
makes this race less likely. It does not address concurrent PUT correctness for
the same resource.

Sequential retries of the same deterministic PUT normally return HTTP `204`.
The failure specifically appeared when two requests for the same identifier
overlapped.

This may happen on other PUT endpoints aswell!

Martin187187:
Closes #565
Closes #566
Closes #570
Closes #590

## Summary

This PR replaces delete-and-recreate `PUT` updates for Asset Administration
Shells and Submodels with transactional, in-place reconciliation. Persisted
rows are compared by their normalized database representation, and only changed
sections are updated. This reduces write amplification, preserves stable row
identities, avoids unnecessary registry synchronization, and fixes duplicate-key
conflicts from concurrent updates to the same resource.

## What changed

- Add reconciliation plans for existing AAS and Submodel resources:
  - update unchanged-type rows in place;
  - insert or delete only added, removed, or type-changed content;
  - match named Submodel Elements by `idShort` and list children by position;
  - match AAS Submodel references by unique Submodel identity when possible and
    preserve them across reordering;
  - skip the live-model write entirely for semantic no-op replacements.
- Execute each live-model reconciliation as one set-wise, parameterized
  PostgreSQL statement generated with Goqu. The reconciliation plan is passed as
  JSON, keeping the SQL shape and parameter count constant as resource size
  grows.
- Lock existing AAS and Submodel root rows for update and reconcile them without
  deleting those roots. This serializes overlapping updates to the same existing
  resource and eliminates the internal duplicate-key `409` race.
- Preserve managed Submodel file references and AAS thumbnails when their owning
  content is unchanged, while cleaning up changed or removed managed content.
- Return structured `IsUpdate`, `Changed`, and previous-state information from
  repository writes so the AAS Environment can avoid unnecessary follow-up
  work.
- Compare derived AAS and Submodel descriptors canonically and synchronize the
  registries only when descriptor-relevant content changed. Referencing AAS
  descriptors are likewise skipped for unchanged Submodel updates.
- Keep authorization, mutation, readback, applicable history/evidence creation,
  and descriptor synchronization in the same transaction. Existing Submodel
  `PUT` checks use the `UPDATE` formula against complete internal snapshots
  without allowing fragment filters to truncate the reconciliation input.
- Preserve API replacement semantics and create/update status behavior.

## Database migration

- Add schema patch `v1.1.13`.
- Make the Submodel Element sibling `idShort` and position uniqueness constraints
  deferrable so set-wise reordering can complete atomically.
- Register the patch in the configuration service and advance the current schema
  version to `v1.1.13`.

## Test coverage

- Add AAS and Submodel integration tests that verify existing database row IDs
  survive full replacements, including a 1,200-element Submodel.
- Add concurrent integration coverage with eight identical `PUT` requests to the
  same existing AAS and Submodel; every request must return `204 No Content`.
- Cover semantic no-op detection, metadata-only and nested-element changes,
  model-type replacement, list and reference reordering, managed thumbnail
  preservation, and canonical `xs:dateTime` comparison.
- Verify constant reconciliation SQL shape and parameter count at scale, exact
  affected-row validation, and single-statement mutation execution.
- Cover pre- and post-update ABAC denial rollback, history behavior, registry
  descriptor comparison, read-query shape, and large-object cleanup regressions.


internal/aasenvironment/custom_aas_repository_service.go
```diff
@@ -514,6 +477,60 @@ func (s *CustomAASRepositoryService) PutSubmodelByIdAasRepository(ctx context.Co
...
 
+func (s *CustomAASRepositoryService) putSubmodelAndSyncDescriptors(
+	ctx context.Context,
+	aasIdentifier string,
+	submodelIdentifier string,
+	submodel types.ISubmodel,
+) (bool, error) {
+	isUpdate := false
+	err := s.ExecuteInTransaction(func(tx *sql.Tx) error {
+		putResult, putErr := s.persistence.SubmodelRepository.PutSubmodelInTransactionWithResult(ctx, tx, submodelIdentifier, submodel)
```
chris-koo: [Medium] Acquire the AAS and Submodel locks in a consistent order. This PUT path locks the Submodel first and then, at `CreateSubmodelReferenceInAssetAdministrationShellInTransaction`, locks the AAS. The delete superpath in the same service does the inverse (AAS reference deletion first, then Submodel deletion). Concurrent PUT and DELETE requests for the same pair can therefore each hold one row/advisory lock while waiting for the other, causing PostgreSQL to abort one transaction as a deadlock and surface a 500. Acquire both resources in one canonical order (for example AAS before Submodel) in every composite operation, ideally through a shared ordered-lock helper, and add a concurrent PUT-vs-DELETE regression test.

internal/aasrepository/persistence/aas_database.go
```diff
@@ -648,12 +658,40 @@ func (s *AssetAdministrationShellDatabase) createSubmodelReferenceInAssetAdminis
...
+			tx, aasDBID, submodelIdentifier, "AASREPO-NEWSMREFINAAS",
+		)
+		if existsErr != nil {
+			return existsErr
+		}
+		if exists {
+			return common.NewErrConflict("AASREPO-NEWSMREFINAAS-CONFLICT Submodel reference to Submodel with ID '" + submodelIdentifier + "' already exists in Asset Administration Shell with ID '" + aasIdentifier + "'")
+		}
+	}
+	previous, err := s.getAssetAdministrationShellMapByDBIDInTransaction(auth.ContextWithoutQueryFilter(ctx), tx, aasDBID)
```
chris-koo: [Medium] Keep single-reference mutations proportional to the changed row. Posting one Submodel reference now reconstructs the complete AAS (including every SpecificAssetId and Submodel reference), diffs the whole object, and executes the general reconciliation statement; the delete-reference path repeats the same pattern. With history evidence disabled, the previous implementation performed only the targeted lookup/inserts, so this changes a constant-size endpoint into O(size of AAS) CPU, allocation, and database transfer and will degrade sharply for large shells. Retain the new AAS lock for serialization but use targeted append/delete SQL for these specialized endpoints, loading a full snapshot only when history evidence requires it; cover the endpoint with a large-reference scale test or benchmark.

chris-koo:
I checked your last two commits. They resolve the problems identified in my earlier review :)

PR #609, opened 2026-08-19
AASX File Server Async Profile

### LINKED ISSUE #531: [Ticket] Add the AASX File Server async profile (SSP-002)
The Go AASX File Server currently implements and advertises the synchronous SSP-001 profile. API 3.2 also defines SSP-002 for asynchronous package uploads, but those endpoints are not available yet.

---

## Rules
- **What:** Implement the AASX File Server SSP-002 endpoints defined by the IDTA API 3.2 specification:
  - `POST /packages-async`
  - `GET /packages-async/status/{handleId}`
  - `GET /packages-async/result/{handleId}`
- **When:** A request accepted by `POST /packages-async` must return `202 Accepted`, an `OperationHandle`, and a `Location` header pointing to the status endpoint. While the operation is running, the status endpoint returns `200`. Once it reaches a terminal state, the status endpoint redirects to the result endpoint as defined by the specification.
- **How:** Use the shared persistent asynchronous-handle lifecycle introduced by #525. Handles and results must not be process-local, must remain owner-scoped, and must follow the same terminal retention, cleanup, and interrupted-worker recovery rules as the other asynchronous APIs.
- The uploaded file must be moved into durable staging before the request returns `202`. The background operation must not depend on a multipart stream or temporary resource owned by the completed HTTP request.
- Reuse the existing upload size limits, multipart validation, PostgreSQL large-object handling, and AAS ID validation.
- Do not buffer complete AASX packages in memory.
- Preserve the existing synchronous SSP-001 endpoints and response formats.
- Add SSP-002 to the service description without removing SSP-001.
- All generated and handwritten API code must remain aligned with the checked-in OpenAPI document.

---

## Entry Points
- `POST /packages-async`
- `GET /packages-async/status/{handleId}`
- `GET /packages-async/result/{handleId}`
- `GET /description`
- `cmd/aasxfileserverservice/openapi.yaml`
- `pkg/aasxfileserverapi/go/`
- `internal/aasxfileserver/api/`
- `internal/aasxfileserver/persistence/`
- `cmd/aasxfileserverservice/main.go`

---

## Acceptance Criteria
- [ ] `POST /packages-async` accepts the multipart request defined by SSP-002 and returns `202`, an `OperationHandle`, and a context-aware status `Location`.
- [ ] A handle created through one service replica can be polled and retrieved through another replica.
- [ ] The status endpoint returns the specified running response and redirects to the result endpoint after completion.
- [ ] The result endpoint returns the SSP-002 `BaseOperationResult` representation for successful and failed uploads.
- [ ] Restarting or losing the worker that accepted an upload produces a deterministic terminal result instead of an unexplained `404`.
- [ ] A staged upload remains available until the asynchronous operation finishes or is recovered, and abandoned staging data is cleaned up.
- [ ] Owner isolation and authorization are preserved across replicas.
- [ ] The service description advertises both SSP-001 and SSP-002.
- [ ] Existing synchronous AASX File Server integration tests continue to pass unchanged.
- [ ] Integration tests cover successful upload, validation failure, cross-replica polling, interrupted-worker recovery, terminal retention, and cleanup.

---

## Risks and Assumptions
- **Risk:** The current upload handler closes staged multipart data when the request finishes. Reusing it unchanged would make the background operation race with cleanup.
- **Risk:** Large packages could be copied unnecessarily if durable staging and final package storage are implemented as separate full writes.
- **Risk:** Interrupted uploads can leak PostgreSQL large objects unless handle cleanup and staging cleanup are coordinated.
- **Assumption:** SSP-001 remains supported alongside SSP-002 rather than being replaced by it.
- **Assumption:** Interrupted work follows the deterministic failure semantics from #525; this ticket does not introduce a general durable job queue.

---

## References and Notes
- [AASX File Server SSP-002, API 3.2](https://github.com/admin-shell-io/aas-specs-api/blob/v3.2.0/AasxFileServerServiceSpecification/V3.2_SSP-002.yaml)
- [Shared API schemas containing `OperationHandle` and `BaseOperationResult`](https://github.com/admin-shell-io/aas-specs-api/blob/v3.2.0/Part2-API-Schemas/openapi.yaml)
- [Current BaSyx Go SSP-001 OpenAPI document](https://github.com/eclipse-basyx/basyx-go-components/blob/main/cmd/aasxfileserverservice/openapi.yaml)
- [Current AASX File Server service implementation](https://github.com/eclipse-basyx/basyx-go-components/tree/main/internal/aasxfileserver)
- Related persistence ticket: #525

---

## Dependencies and Blockers
- **Dependency:** #525 should provide the shared PostgreSQL-backed asynchronous handle store, atomic state transitions, terminal retention, and cleanup behavior.
- **Blockers:** None beyond agreeing on and merging the reusable persistence API from #525.

chris-koo:
Implements the AASX File Server async profile with endpoints for starting uploads, checking execution status, and retrieving results. Adds persistent async job handling, database migrations, authorization support, updated OpenAPI-generated models and routes, plus comprehensive unit, contract, and integration tests.

Closes #531


FriedJannik (approved):
LGTM

cmd/aasxfileserverservice/main.go
```diff
@@ -136,6 +188,28 @@ func runServer(ctx context.Context, configPath string) error {
...
+			minimumAASXAsyncDatabaseHeadroom,
+		)
+	}
+	databaseHeadroom := max(maximumOpenConnections/aasxAsyncDatabaseHeadroomDivisor, minimumAASXAsyncDatabaseHeadroom)
+	return maximumOpenConnections - databaseHeadroom, nil
+}
+
+func requireAsyncAuthentication(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
+		authorization := strings.TrimSpace(request.Header.Get("Authorization"))
```
aaronzi: This check can be bypassed with a non-breaking space before `Bearer`, e.g. `Authorization: \u00a0Bearer junk`. OIDC checks the untrimmed header and treats the request as anonymous, while this check trims the whitespace and lets it through. With an ABAC rule allowing anonymous access to the route, the request reaches the upload handler without a valid token. I reproduced this over HTTP: `Bearer junk` returns 401, but the NBSP-prefixed value reaches the handler with `owner=anonymous`.

I'd check the verified authentication state in the request context here instead of parsing the header again, and add a regression test covering the full middleware chain.

pkg/aasxfileserverapi/go/api_aasx_async_file_server_status_api.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// GetAasxAsyncStatus returns or redirects the status of an asynchronous package upload.
+func (controller *AASXAsyncFileServerStatusAPIAPIController) GetAasxAsyncStatus(writer http.ResponseWriter, request *http.Request) {
+	handleID := chi.URLParam(request, "handleId")
+	if handleID == "" {
+		controller.errorHandler(writer, request, &RequiredError{"handleId"}, nil)
+		return
+	}
+	result, err := controller.service.GetAasxAsyncStatus(request.Context(), handleID)
```
aaronzi, line 76: The referenced [HandleId parameter](https://github.com/admin-shell-io/aas-specs-api/blob/main/Part2-API-Schemas/openapi.yaml) requires UTF8-BASE64-URL encoding, but we pass the path value directly into the handle lookup. The result endpoint does the same, and the Location headers only use `url.PathEscape`. A client encoding the returned handle as specified therefore gets 404. I checked both endpoints: the raw handle works, while its Base64URL-encoded form returns 404.

I'd decode the incoming path handle before lookup and Base64URL-encode it when building the status/result Location headers. We should also cover encoded handles in the contract tests.

cmd/aasxfileserverservice/main.go
```diff
@@ -96,12 +107,41 @@ func runServer(ctx context.Context, configPath string) error {
 	}
 	slog.InfoContext(ctx, "PostgreSQL connection established")
 
-	aasxSvc := aasxapi.NewAASXFileServerAPIAPIService(aasxDatabase)
+	asyncExecutionCapacity, err := aasxAsyncExecutionCapacity(sharedDB.Stats().MaxOpenConnections)
```
aaronzi: This makes `postgres.maxOpenConnections: 1` fail startup, even though it's still a valid configuration and previously worked for synchronous uploads. Since async initialization is unconditional, existing SSP-001-only deployments using a single connection now lose the entire service on upgrade.

I'd preserve synchronous startup when the pool has insufficient async capacity, either by disabling async admission in that case or by providing an explicit option to disable the async profile.

pkg/aasxfileserverapi/go/model_execution_state.go
```diff
@@ -0,0 +1,85 @@
...
+*
+* SPDX-License-Identifier: MIT
+******************************************************************************/
+
+package openapi
+
+import "fmt"
+
+// ExecutionState identifies the lifecycle state of an asynchronous operation.
+type ExecutionState string
```
aaronzi: Small DRY cleanup: we already have the same six states, validation and parser in [the shared ExecutionState model](https://github.com/eclipse-basyx/basyx-go-components/blob/eb68cf6a607df90f5d66cb3056c07070bc165da6/internal/common/model/model_execution_state.go#L42-L100). I'd use `type ExecutionState = model.ExecutionState` here and forward the constants/helper functions so callers can keep using the current package API. We already use this alias pattern for [SubmodelDescriptor](https://github.com/eclipse-basyx/basyx-go-components/blob/eb68cf6a607df90f5d66cb3056c07070bc165da6/pkg/smregistry/api_submodel_registry_api.go#L34).

I tried this locally and it works. The [contract assertions](https://github.com/eclipse-basyx/basyx-go-components/blob/eb68cf6a607df90f5d66cb3056c07070bc165da6/cmd/aasxfileserverservice/openapi_async_contract_test.go#L193-L201) also need to check the underlying type and enum values instead of requiring a local string declaration; with that adjustment, the affected unit/contract tests and AASX integration suite pass.

internal/aasxfileserver/persistence/async_upload.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+	if err := upload.closeReaderLocked(); err != nil {
+		return err
+	}
+	if upload.promoted {
+		upload.closed = true
+		return nil
+	}
+
+	ctx, cancel := context.WithTimeout(context.WithoutCancel(upload.ctx), 5*time.Second)
```
aaronzi: I'd reuse [asyncjob.NewPersistenceContext](https://github.com/eclipse-basyx/basyx-go-components/blob/eb68cf6a607df90f5d66cb3056c07070bc165da6/internal/common/asyncjob/manager.go#L301-L305) here:

```go
ctx, cancel := asyncjob.NewPersistenceContext(upload.ctx)
```

It already detaches cancellation and applies the same five-second timeout. This keeps upload cleanup on the same persistence timeout policy as [async failure recording](https://github.com/eclipse-basyx/basyx-go-components/blob/eb68cf6a607df90f5d66cb3056c07070bc165da6/internal/aasxfileserver/api/api_aasx_file_server_api_service.go#L363), and the package is already imported. I checked the replacement locally; the persistence tests and AASX integration suite pass.

internal/common/binarycontent/staged_large_object.go
```diff
@@ -236,6 +236,14 @@ type transactionLargeObjectReader struct {
 	mu         sync.Mutex
 }
 
+// ReadSeekCloser combines the operations required for independently reopened
+// PostgreSQL large objects.
+type ReadSeekCloser interface {
```
aaronzi: Small cleanup: this is exactly the standard library's [io.ReadSeekCloser](https://pkg.go.dev/io#ReadSeekCloser). I'd use that directly, or keep the current name as `type ReadSeekCloser = io.ReadSeekCloser` if we want to preserve the existing references. There are no additional methods or constraints here that need a separate interface. I checked the alias locally; the binary-content tests and AASX integration suite pass.

aaronzi (approved):

PR #643, opened 2026-08-30
Remove redundant SubmodelElement path index

aaronzi:
## What changed

- remove `ix_sme_sub_path(submodel_id, idshort_path)` through schema patch `v1.1.17`
- retain `ix_sme_sub_path_id(submodel_id, idshort_path, id)`, which covers the same lookup prefix and the stable path ordering used by bounded Submodel reads
- keep the existing batched Submodel creation implementation unchanged

## Why

The representative HARTING create workload persists about 30 SubmodelElements per request. PostgreSQL was maintaining both path indexes for every inserted element even though the newer index has the same leading columns.

A transactional planner check showed that exact `submodel_id + idshort_path` lookups move directly to `ix_sme_sub_path_id` when the duplicate index is absent. Ordered page reads already use that index.

The earlier attempt to replace the insert batch with one JSON reconciliation statement was rejected after cluster testing because it did not improve the throughput ceiling and increased tail latency. That implementation is not part of the final diff.

## Validation

- `go test ./internal/basyxconfigurationservice/... ./cmd/basyxconfigurationservice/...`
- `go vet ./cmd/basyxconfigurationservice/... ./internal/basyxconfigurationservice/...`
- `go clean -testcache && go test -v ./internal/submodelrepository/integration_tests`
- full GitHub Actions matrix, including integration, security, query, example, lint, vet, formatting, changelog and database-patch checks
- all checks passed

### Cluster evidence

Four Submodel Repository replicas, two PostgreSQL replicas and representative HARTING payloads averaging about 30 SubmodelElements. The removed index occupied 9,455 MB on the benchmark database.

| CreateSubmodel setup | Concurrency | Duration | RPS | p50 | p95 | Failures |
|---|---:|---:|---:|---:|---:|---:|
| BaSyx Go 1.0.10 baseline | 80 | 300 s | 653.7 | 102.4 ms | 241.9 ms | 0 |
| PR 643 | 80 | 60 s | 670.9 | 85.4 ms | 251.3 ms | 0 |

The final create result is 2.6% above the release baseline. The primary benefit is removing one redundant B-tree update for every persisted SubmodelElement rather than claiming a large endpoint-throughput change.

Read controls at their existing concurrency points showed no significant regression:

| Endpoint | BaSyx Go 1.0.10 | PR 643 | Change | Failures |
|---|---:|---:|---:|---:|
| ReadSubmodelById | 4,936.8 RPS | 4,952.5 RPS | +0.3% | 0 |
| GetSubmodelElementByIdShortPath | 8,502.9 RPS | 8,664.2 RPS | +1.9% | 0 |
| GetSubmodelElementValueOnlyByIdShortPath | 8,517.1 RPS | 8,839.0 RPS | +3.8% | 0 |
| GetAllSubmodels?limit=100 | 153.9 RPS | 147.4 RPS | -4.2% | 0 |

The representative GetAll result remains within the normal run variance threshold. Its bounded page query already used `ix_sme_sub_path_id` before this change. Exact path lookups also move to that retained index with the same `(submodel_id, idshort_path)` lookup prefix.

### DIFF
--- a/database/patches/1_1_17.sql
+++ b/database/patches/1_1_17.sql
@@ -1,0 +1,38 @@
+/*******************************************************************************
+* Copyright (C) 2026 the Eclipse BaSyx Authors and Fraunhofer IESE
+*
+* Permission is hereby granted, free of charge, to any person obtaining
+* a copy of this software and associated documentation files (the
+* "Software"), to deal in the Software without restriction, including
+* without limitation the rights to use, copy, modify, merge, publish,
+* distribute, sublicense, and/or sell copies of the Software, and to
+* permit persons to whom the Software is furnished to do so, subject to
+* the following conditions:
+*
+* The above copyright notice and this permission notice shall be
+* included in all copies or substantial portions of the Software.
+*
+* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+* MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+* NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+* LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+* OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+* WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*
+* SPDX-License-Identifier: MIT
+******************************************************************************/
+
+-- ix_sme_sub_path_id has the same leading columns and covers every lookup
+-- previously served by ix_sme_sub_path without maintaining a second path index.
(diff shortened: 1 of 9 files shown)


database/patches/1_1_17.sql
```diff
@@ -0,0 +1,38 @@
...
+* LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+* OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+* WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*
+* SPDX-License-Identifier: MIT
+******************************************************************************/
+
+-- ix_sme_sub_path_id has the same leading columns and covers every lookup
+-- previously served by ix_sme_sub_path without maintaining a second path index.
+DROP INDEX IF EXISTS ix_sme_sub_path;
```
aaronzi, line 28: The removed index returns after the next configuration-service execution. The service always uploads `base.sql`, which recreates `ix_sme_sub_path`. Patch 1.1.17 then skips because the database already reports v1.1.17. This silently restores the roughly 9.5 GB index and its write amplification.

aaronzi: Will be addressed by this PR that has to be merged first: https://github.com/eclipse-basyx/basyx-go-components/pull/644

Issue #188, opened 2026-02-23
[Ticket] Eventing for CUD Operations


aaronzi:
## User Story
> **As a** BaSyx user/developer,  
> **I want** automatic MQTT events whenever resources are created, updated, or deleted across BaSyx Go components,  
> **so that** external systems can reliably react to model lifecycle changes in near real time.

## Rules
- **What:**
  - Emit events for all CUD operations (Create/Update/Delete), never for read operations.
  - In scope components:
    - AAS Registry
    - Submodel Registry
    - Discovery
    - AAS Repository
    - Submodel Repository
    - Concept Description Repository
    - Registry of Infrastructures
  - Use MQTT as transport for MVP.
  - Use a common event envelope across components:
    - `eventType`
    - `component`
    - `operation`
    - `identifiers`
    - `timestamp`
    - `correlationId`
  - Use hierarchical topics: `basyx/<component>/<resource>/<operation>`.
  - Publish full affected resource JSON for create/update (with envelope metadata).
  - Default MQTT semantics:
    - QoS = 1 (configurable)
    - retained = false (configurable)
  - Security/authorization for eventing is out of scope.
- **When:**
  - Trigger event publish only after successful persistence/commit of the CUD operation.
  - If publish fails, API response remains successful (non-blocking).
- **How:**
  - Implement shared/core eventing logic in common code.
  - Prefer decorator/wrapper pattern for publish integration (aligned with BaSyx Java approach).
  - Add service configuration via env vars and config files.
  - Eventing is disabled by default and enabled via `MQTT_ENABLED=true`.

## Entry Points
- Shared descriptor write paths:
  - `internal/common/descriptors/AssetAdminShellDescriptorHandler.go`
  - `internal/common/descriptors/SubModelDescriptorHandler.go`
  - `internal/common/descriptors/InfrastructureDescriptorHandler.go`
  - `internal/common/descriptors/DescriptorTransactionHelper.go`
- Submodel Repository write paths:
  - `internal/submodelrepository/persistence/submodel_database.go`
- API service entry points:
  - `internal/aasregistry/api/api_asset_administration_shell_registry_api_service.go`
  - `internal/smregistry/api/api_submodel_registry_api_service.go`
  - `internal/discoveryservice/api/api_asset_administration_shell_basic_discovery_api_service.go`
  - `internal/conceptdescriptionrepository/api/api_concept_description_repository_api_service.go`
[...]

PR #647, opened 2026-09-01
Feature: AAS Event Feed API

tfunken:
This pull request implements the AAS event feed as an experimental feature according to the [AAS Eventing Spec](https://docs.researchandprototyping.com/docs/specifications/api/eventing-event-feed.html).

Scope:
- CloudEvents Event Feed: GET /events for aasenvironemnet, aasrespository and submodelrepository
- Created / updated / deleted for AAS, Submodel, and Asset (when globalAssetId is set) 
- PCN events (io.admin-shell.pcn.v1)
- DB patch v1.2.0 (feed_events), config under eventing.feed 
- Sample configs: eventing.enabled: true for AAS Repository, Submodel Repository, AAS Environment 

Out of scope: 
- MQTT / Kafka sinks and outbox
- Security for event data payload
- Filtering event data field


cmd/aasrepositoryservice/config.yaml
```diff
@@ -74,6 +74,22 @@ general:
 #   privateKeyPath: "./rsa-key.pem"
 #   certificateChainPath: "./rsa-cert.pem"
 
+# Event Feed (CloudEvents REST). Requires DB schema >= v1.2.0 (feed_events).
+# Set enabled: true for local manual testing.
+eventing:
+  enabled: true
```
aaronzi: **High — make the experimental feed opt-in.** This checked-in config is copied into the release image, and the same setting is enabled in the Submodel Repository and AAS Environment configs. An upgrade would therefore add `/events` without an operator explicitly enabling it, exposing identifiers, relationships, semantic IDs, and mutation timing. Please ship all standard configs with `eventing.enabled: false` and `publicAccess: false`, and put enabled values in a separate manual-test/example configuration. Please also test that the route and OpenAPI operation are absent by default.

internal/common/eventfeed/service.go
```diff
@@ -0,0 +1,263 @@
...
+
+func (s *Service) Write(ctx context.Context, event FeedEvent) error {
+	if s == nil || s.repo == nil {
+		return nil
+	}
+	return s.repo.Save(ctx, event)
+}
+
+// PublishBestEffort writes event and logs build or persistence failures.
+func (s *Service) PublishBestEffort(ctx context.Context, event FeedEvent, err error) {
```
aaronzi: **High — do not publish post-commit on a best-effort basis.** The model transaction has already committed when callers reach this method, while build/insert failures and a cancelled request context are only logged. That permanently loses events with no retry path, and the paired AAS/asset events can be only partially written. This also conflicts with the transaction-scoped outbox direction documented in `docu/developer/aas_v3_2_runtime.md` and #188. Please enqueue the normalized mutation/feed record in the same database transaction as the model change and materialize it idempotently from committed state.

internal/common/eventfeed/builder.go
```diff
@@ -0,0 +1,281 @@
...
+		"semanticId": semRef,
+	}
+	return b.build(eventType, submodelID, sourceSuffixSubmodel, schemaSubmodelFull, schemaSubmodelCompact, full, compact)
+}
+
+func (b *Builder) build(
+	eventType, subject, sourceSuffix, schemaFull, schemaCompact string,
+	dataFull, dataCompact map[string]any,
+) (FeedEvent, error) {
+	now := b.now()
```
aaronzi, line 212: **High — application time is not a safe feed ordering/cursor key.** The timestamp is allocated before a separate database insert. If A receives T1 and stalls while B receives T2, inserts, and is consumed, A can later insert below the already-issued `(time,id)` cursor and be skipped forever. Multiple instances and clock skew make this worse, and per-entity commit order can be reversed. Please use a durable database-backed ordering key produced by the transactional mutation/outbox path and key cursors on that ordering rather than wall-clock time.

internal/aasrepository/api/api_asset_administration_shell_repository_api_service.go
```diff
@@ -463,6 +488,17 @@ func (s *AssetAdministrationShellRepositoryAPIAPIService) DeleteAssetAdministrat
 		return newAPIErrorResponse(decodeErr, http.StatusBadRequest, operation, "MalformedAssetAdministrationShellIdentifier"), nil
 	}
 
+	aas, getErr := s.assetAdministrationShellBackend.GetAssetAdministrationShellByID(ctx, decodedIdentifier)
```
aaronzi: **High — this pre-delete read changes established security and consistency behavior even when eventing is disabled.** The existing security integration test now returns 404 instead of the expected 403 for an unauthorized delete. With a reader replica, a recently created AAS can also be falsely reported missing; independently, this snapshot can race the writer delete. Please read and lock the pre-delete state inside the same writer transaction as deletion. Using writer reads alone would address replica lag but not the race. The standalone Submodel delete has the same issue.

internal/common/eventfeed/builder.go
```diff
@@ -0,0 +1,281 @@
...
+			continue
+		}
+		out = append(out, map[string]any{"referredSemanticId": externalReference(semanticID)})
+	}
+	return out
+}
+
+func modelReference(keyType, value string) map[string]any {
+	return map[string]any{
+		"type": "ModelReference",
```
aaronzi, line 265: **High — the generated payload does not validate against the schema advertised in `dataschema`.** The linked draft's Reference and Key definitions require the discriminator `_type` and disallow additional properties, while this emits `type`. `externalReference("")` also emits an empty `keys` array although the schema requires at least one entry. Please use the established metamodel JSON serialization or produce the exact draft representation, define valid behavior for a missing semantic ID, and validate every generated REGULAR/COMPACT payload against its advertised JSON Schema in tests.

internal/common/eventfeed/service.go
```diff
@@ -0,0 +1,263 @@
...
+			SupportedPrefixes: []string{"rsql"},
+			RSQL:              RSQLCapabilities{Operators: []string{"==", "!=", "in", "out"}},
+		},
+		Presentation: PresentationCapabilities{
+			Supported: []string{string(PresentationFull), string(PresentationCompact)},
+			Default:   string(PresentationFull),
+		},
+		MaxAge:      s.cfg.MaxAgePeriod(),
+		MaxPageSize: s.cfg.MaxPageSize,
+		Auth: AuthCapabilities{
```
aaronzi: **High — these auth settings are descriptive only and create a false security expectation.** Neither `publicAccess` nor `bearerAuth` affects middleware or the `/events` handler. With ABAC disabled, `publicAccess: false`/`bearerAuth: true` still leaves the endpoint anonymous; with ABAC enabled, `/events` has no rights mapping and is denied for every policy. Even a generic READ mapping would still need event-level authorization so a caller cannot see events for models it cannot read. Please either remove these knobs and document inheritance of established service security, or implement and integration-test the advertised modes and ABAC filtering.

internal/common/eventfeed/handler.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+const httpComponent = "EVENTFEED"
+
+// RegisterRoutes registers the enabled event feed endpoints on r.
+func RegisterRoutes(r chi.Router, svc *Service) {
+	if r == nil || svc == nil || !svc.cfg.Enabled {
+		return
+	}
+	r.Get("/events", svc.handleGetEvents)
```
aaronzi, line 47: **Medium — the required capabilities endpoint is missing.** `Service.Capabilities()` is implemented but unreachable because only `/events` is registered. The normative draft section requires `/.well-known/event-feed.json`, which clients use for API/event versions, schemas, filters, presentation modes, page limits, and auth discovery. Please register it at the specified location, include it in the conditional OpenAPI document, and test that the response stays consistent with runtime behavior.

internal/common/eventfeed/types.go
```diff
@@ -0,0 +1,157 @@
...
+package eventfeed
+
+import "time"
+
+// Presentation selects which stored payload variant is returned to consumers.
+type Presentation string
+
+const (
+	// PresentationFull returns the full event data representation.
+	PresentationFull Presentation = "FULL"
```
aaronzi, line 37: **Medium — this public value differs from the referenced draft.** The normative presentation values are `REGULAR` and `COMPACT`; a conforming client currently gets HTTP 400 for `presentation=REGULAR`. Please use `REGULAR` on the wire and in schema-capability keys. If pre-release compatibility is necessary, `FULL` could temporarily be accepted as an undocumented/deprecated alias. The capabilities operator names should likewise advertise `=in=` and `=out=`, not `in` and `out`.

internal/common/eventfeed/handler.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func (s *Service) handleGetEvents(w http.ResponseWriter, r *http.Request) {
+	q := r.URL.Query()
+	presentation := Presentation(strings.ToUpper(strings.TrimSpace(q.Get("presentation"))))
+	if presentation == "" {
+		presentation = PresentationFull
+	}
+	limit := 100
```
aaronzi: **Medium — derive the omitted limit from `s.cfg.MaxPageSize`.** A valid deployment with `maxPageSize: 50` currently returns 400 for plain `GET /events`, because the handler injects 100 and validation rejects it. The draft also requires an omitted limit to use the advertised `maxPageSize`. Please add a handler test with a configured maximum below 100 and avoid a hard-coded OpenAPI default that can disagree with runtime configuration.

internal/submodelrepository/api/api_submodel_repository_api_service.go
```diff
@@ -1308,7 +1344,9 @@ func (s *SubmodelRepositoryAPIAPIService) PutSubmodelByID(ctx context.Context, s
 		return newAPIErrorResponse(err, http.StatusInternalServerError, operation, "InternalServerError"), nil
 	}
 
-	if isUpdate {
+	publishSubmodelEvent(ctx, s.eventFeed, s.submodelBackend, !putResult.IsUpdate, submodel, putResult.Previous)
```
aaronzi: **Medium — this publishes an update even when persistence reports `putResult.Changed == false`.** That creates `submodel.updated` records for no-op PUTs and disagrees with the existing reconciliation/history result. Please only publish for an actual persisted change. More broadly, placing the hook here is why PATCH/SME/value/file/internal writes are missed; the event should be derived from the shared transaction-scoped mutation record instead of individual HTTP handlers.

internal/aasrepository/api/eventfeed_hooks.go
```diff
@@ -0,0 +1,163 @@
...
+		return nil
+	}
+	aasIDs, err := backend.ListAASIdentifiersBySubmodelID(ctx, submodelID)
+	if err != nil || len(aasIDs) == 0 {
+		return nil
+	}
+	seen := make(map[string]struct{}, len(aasIDs))
+	globalAssetIDs := make([]string, 0, len(aasIDs))
+	for _, aasID := range aasIDs {
+		aas, getErr := backend.GetAssetAdministrationShellByID(ctx, aasID)
```
aaronzi: **Medium — this enrichment is an N+1 full-model read path and silently degrades event data.** One mutation can list all referencing AAS IDs and then hydrate every AAS sequentially; the semantic-ID helper similarly loads every referenced Submodel. Errors are silently skipped, so emitted events can be incomplete. Some callers also evaluate enrichment before the helper checks whether eventing is enabled. Please guard before doing work and add bulk GOQU projections/joins for only the required IDs, or derive the fields from the atomic mutation snapshot/outbox and surface failures.

internal/common/eventfeed/repository.go
```diff
@@ -0,0 +1,205 @@
...
+	if err != nil {
+		return FeedEvent{}, false, fmt.Errorf("EVENTFEED-FINDBYID-SCAN: %w", err)
+	}
+	return e, true, nil
+}
+
+// FindPage returns one ordered page of retained events.
+func (r *Repository) FindPage(ctx context.Context, q domainQuery, presentation Presentation) ([]FeedEvent, error) {
+	ds := r.dialect.From("feed_events").
+		Select("id", "event_type", "subject", "source", "time",
```
aaronzi: **Medium — this query always loads both JSONB payloads and both schema variants although only one presentation is returned.** For large PCN payloads that makes COMPACT reads unnecessarily expensive. In addition, all four advertised filter columns are unindexed; `(time,id)` alone can require scanning most of the retention window before filling a filtered page. Please project only the selected payload/schema and validate expected filter plans with `EXPLAIN`, adding composite indexes such as `(event_type,time,id)` where justified. The standalone `time` index in the migration is likely redundant with `(time,id)`.

internal/common/eventfeed/publisher.go
```diff
@@ -0,0 +1,186 @@
...
+func (m *Module) StartRetentionLoop(ctx context.Context) {
+	if m == nil || !m.Enabled() {
+		return
+	}
+	interval := m.cfg.CleanupInterval
+	if interval <= 0 {
+		interval = 24 * time.Hour
+	}
+	go func() {
+		ticker := time.NewTicker(interval)
```
aaronzi, line 91: **Medium — cleanup should not wait for the first ticker.** Deployments restarted more frequently than this interval may never clean up. Once a backlog exists, the current unbounded DELETE runs as one potentially large transaction, creating WAL/vacuum and locking pressure; multiple service replicas can also run it simultaneously. Please run a cleanup pass at startup, delete in bounded batches, and coordinate workers with an advisory lock or designated instance.

internal/common/configuration.go
```diff
@@ -1032,12 +1057,61 @@ func validateIntegrityAnchorConfig(cfg HistoryIntegrityAnchorConfig) error {
...
+	}
+	sourceBaseURL = strings.TrimSpace(cfg.Feed.SourceBaseURL)
+	if sourceBaseURL == "" {
+		sourceBaseURL = "http://localhost"
+	}
+	schemaBaseURL = strings.TrimSpace(cfg.Feed.SchemaBaseURL)
+	if schemaBaseURL == "" {
+		schemaBaseURL = "https://admin-shell.io/events/schemas"
+	}
+	return maxAgeDays, hardDeleteGraceDays, maxPageSize, cleanupHours, sourceBaseURL, schemaBaseURL, cfg.Feed.PublicAccess, cfg.Feed.BearerAuth, cfg.Enabled
```
aaronzi: **Low — give the REST feed its own enable switch.** Mapping global `eventing.enabled` directly to the feed means that implementing and enabling a future MQTT/Kafka/outbox sink would also expose `/events`. Please add `eventing.feed.enabled` (optionally underneath a global master switch) so future event transports can be enabled independently without changing the service's HTTP exposure.

aaronzi (changes requested):
Thanks for contributing this! Functionality-wise this already looks very promising, especially the dedicated event-feed package, the versioned event types, and the conditional Swagger integration. In the review we had a few findings/remarks; please have a look at them down below. We are happy to support if needed, particularly around integrating this with the existing mutation/history architecture.

The main cross-cutting concern is event capture. The current API-layer hooks cover selected top-level POST/PUT/DELETE operations, but successful PATCH, metadata/value-only, SubmodelElement, file, AssetInformation, thumbnail, reference-deletion, AASX upload, and other internal persistence paths do not emit events. Adding a PCN record through the normal SubmodelElement endpoint therefore does not produce the PCN event. This is the bypass risk discussed in #188. Please move event creation behind the authoritative transaction-scoped mutation path and reuse the existing normalized `history.ChangeEvent`/outbox direction rather than reconstructing mutations after commit in several handlers. This will also make future event-schema versions much easier to add without changing the public feed API.

We also need service/PostgreSQL integration coverage for the default-disabled route and Swagger behavior, all representative mutation paths (including PCN/SME/file/import), transaction failure and request cancellation, ABAC behavior, real filter/pagination SQL, concurrent ordering, and retention. The full test run currently has a PR-related regression in the AAS Repository security suite: `DELETE invisible existing aas - editor denied` expects 403 but now receives 404 due to the new pre-delete lookup.

Finally, the user and developer documentation still describe eventing as a future placeholder that fails fast. Please add an experimental Event Feed page and an opt-in example setup covering security limitations, configuration/environment variables, endpoint and capabilities discovery, delivery guarantees, retention, schema/event versioning, database migration, and compatibility expectations.

The inline comments contain the concrete findings and suggested changes. Once the transaction/security foundation and the missing mutation/test coverage are addressed, the overall shape should be in a good position for another review.

internal/common/eventfeed/service.go
```diff
@@ -0,0 +1,314 @@
...
+			if len(out) == limit {
+				return out, true, nil
+			}
+			out = append(out, event)
+		}
+		if !rawHasMore {
+			break
+		}
+	}
+	return out, hasMore, nil
```
aaronzi: **Medium — do not report end of feed when the authorization scan hits its cap.** After 32 rounds this returns `hasMore=false`, even if more database rows exist. With `limit=1`, 32 hidden events followed by one readable event produce an empty response and no cursor; retrying starts at the same hidden prefix, so the readable event remains unreachable. I reproduced this with the service. Please preserve a continuation position when the scan budget is exhausted, including for empty authorized pages, or apply authorization during selection so pagination can reach the remaining records.

internal/common/security/event_feed_authorizer.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	path := eventReadPath(eventType, subject)
+	if path == "" {
+		return false
+	}
+	opts := grammar.DefaultSimplifyOptions()
+	opts.EnableImplicitCasts = a.Settings.EnableImplicitCasts
+	evaluation := model.AuthorizeWithFilterWithOptions(EvalInput{
+		Method:    httpMethodGet,
+		Path:      path,
```
aaronzi, line 72: **Medium — include the configured context path in the synthetic read request.** `eventReadPath()` returns `/shells/{id}` or `/submodels/{id}`, but IDENTIFIABLE policy mappings include the service base path. With `server.contextPath: /api/v3`, the editor policy authorizes `/api/v3/shells/{id}` while this check denies the same AAS event, leaving the feed empty for an authorized caller. I reproduced this with the existing access rules. Please join the access model's base path to the synthetic request path and cover prefixed AAS and Submodel routes in the tests.

internal/common/eventfeed/submodel_helpers.go
```diff
@@ -0,0 +1,111 @@
...
+	for i, record := range currentRecords {
+		if record == nil {
+			continue
+		}
+		if previous != nil {
+			if idShort := record.IDShort(); idShort != nil && *idShort != "" {
+				if _, exists := previousIDShorts[*idShort]; exists {
+					continue
+				}
+			} else if i < previousCount {
```
aaronzi: **Medium — list position does not identify an existing PCN record.** Children of a SubmodelElementList have no `idShort`, so this skips every position below the previous length. A PUT changing `Records` from `[CN1]` to `[CN2, CN1]` therefore misses the new CN2 notification and emits CN1 again. I reproduced this with the helper: it returns the record whose `ManufacturerChangeID` is CN1. Please identify existing records using stable record identity or compare the actual records, and cover insertion/reordering in the tests.

internal/common/eventfeed/filter.go
```diff
@@ -0,0 +1,182 @@
...
+			if idx < 0 {
+				parts = append(parts, expr[start:])
+				break
+			}
+			parts = append(parts, expr[start:start+idx])
+			start = start + idx + len(" and ")
+		}
+		return parts
+	}
+	return strings.Split(expr, ";")
```
aaronzi: **Medium — respect quoted values when splitting RSQL expressions.** `rsql:event.subject=='urn:example:sm;v1'` is a valid equality filter, but this splits the quoted identifier and returns HTTP 400. The comma split in `parseRSQLValues()` likewise turns `event.subject=in=('urn:example:sm,a','urn:example:sm,b')` into four broken values, so neither identifier matches. Please use parsing that respects quotes, escapes, and grouping, and add regression tests for delimiters inside quoted identifiers.

aaronzi (changes requested):
I rechecked the updated revision and resolved nine earlier threads. Five remain open, with the current evidence added to each: commit ordering, payload schema validation, event-level authorization, no-op PUTs, and retention lock ownership.

There are four additional inline findings covering authorized pagination, nonempty context paths, PCN list changes, and quoted RSQL values. Please address these together with the remaining threads before merging.

Validation: the affected unit/API/persistence packages and the complete Submodel Repository, AAS Repository, AAS Environment, and AAS Repository security integration suites passed after clearing the test cache. Separate reproductions confirmed the issues described in the comments, including the PostgreSQL ordering/locking interleavings and schema validation failures. The existing integration tests still do not cover these failing cases.

The changelog check is also failing because this PR has no unreleased fragment. Please add a `.changes/unreleased` entry for #647 describing the experimental API and its configuration/migration requirements.

internal/common/eventfeed/builder.go
aaronzi: **High — the sequence still does not follow commit order.** I rechecked this on PostgreSQL: transaction A inserts seq=1 and stays open; B inserts seq=2 and commits; the consumer reads B; A then commits. Resuming with `lastEventId=B` applies `seq > 2` and permanently misses A. `BIGSERIAL` assigns values before commit, so moving the insert into the model transaction does not close this race. Please use an ordering/publication mechanism that cannot expose a position before all earlier positions are committed or excluded, and add a two-transaction integration test for this interleaving.

This thread remains open and needs to be resolved before merging.

internal/common/eventfeed/builder.go
aaronzi: **High — schema validation still fails with the current draft.** I validated actual builder output against the JSON Schemas in the linked specification. AAS REGULAR events with submodels fail because `submodelReference()` emits `ExternalReference` where `ModelReference` is required; Asset REGULAR uses the same helper. Submodel events without a semantic ID fail in both REGULAR and COMPACT because `semanticId` is required but omitted. The current draft now uses `type` in most definitions, so the earlier blanket `_type` explanation is no longer the main issue. Please emit the required reference type, agree on a valid schema representation for absent semantic IDs, and validate generated payloads against the actual advertised schemas rather than hand-written shape checks.

This thread remains open and needs to be resolved before merging.

internal/common/eventfeed/service.go
aaronzi: **High — event authorization still skips the model restrictions.** `EventRecordAuthorizer.Allow()` returns `evaluation.Allowed` and discards `evaluation.QueryFilter`. In this engine, `Allowed` can mean that access is conditional on the backend applying that filter. With the existing `viewer_visible_or_key` policy, I reproduced the authorizer returning true for an invisible AAS while the asset-type/specific-asset-ID restrictions remain in `QueryFilter`. `/events` can therefore expose records for models the caller cannot read, contrary to the documented fail-closed behavior. Please enforce the complete filter against authoritative state/snapshots, or deny when the remaining restriction cannot be evaluated, and add a feed integration test using the existing invisible AAS fixture.

This thread remains open and needs to be resolved before merging.

internal/submodelrepository/api/api_submodel_repository_api_service.go
aaronzi: **Medium — no-op PUTs still emit updates through the new shared hook.** `reconcileExistingSubmodelForPutTx()` calls `appendAcknowledgedSubmodelPutHistoryTx()` even when `Changed` is false, and `AppendVersionTx()` now forwards that acknowledgement to the event sink. `handleSubmodel()` inserts an update without comparing the snapshots; AAS PUT has the same issue. I also confirmed that identical previous/current AAS snapshots still issue an `INSERT INTO feed_events`. Please distinguish acknowledged writes from actual content changes at the sink boundary, preserve the intended history behavior, and test that repeated identical AAS/Submodel PUTs leave the feed unchanged.

This thread remains open and needs to be resolved before merging.

internal/common/eventfeed/publisher.go
aaronzi: **Medium — retention lock ownership is still unsafe with the connection pool.** Startup cleanup and bounded deletes are now present, but `TryRetentionLock()` acquires a PostgreSQL session lock through `sql.DB`, and `ReleaseRetentionLock()` can run on a different connection. I reproduced this by borrowing the lock-owning connection for another request before release: the unlock used another session and a second worker still could not acquire the lock afterward. Please pin one `sql.Conn` for lock acquisition and release, check the unlock result, and ensure cancellation does not leave a locked session in the pool.

This thread remains open and needs to be resolved before merging.

internal/common/eventfeed/builder.go
tfunken: Fixed by decoupling commit from publication. feed_events keeps its internal seq (still allocated pre-commit, no longer used for ordering) and gains a new publish_seq, assigned by a periodic background job that only picks up rows with publish_seq IS NULL. Since that selection runs under normal MVCC, it only ever sees already-committed rows — a still-open transaction is invisible to it, so a publish_seq can never be handed out ahead of an earlier-committing transaction nobody has seen yet. cursor and lastEventId are both based on publish_seq now, not seq. Thus, you will never miss an event.

Verified with a two-transaction integration test against real Postgres.

Side effects:

- Delivery latency: an event isn't visible via GET /events until the next assignment tick (bounded by publishIntervalMillis), not immediately on commit.
- lastEventId gating: resuming from an event that exists but hasn't been assigned a publish_seq yet now returns EVENTFEED-QUERY-LASTEVENT-PENDING instead of silently resuming from an unsafe position.
- since is unaffected and remains a best-effort filter only, not a resumable/gap-safe position — same as before.

internal/submodelrepository/api/api_submodel_repository_api_service.go
tfunken: Added Mutation.Acknowledged/AppendAcknowledgedVersionTx, and the no-live-mutation branches for both submodel and AAS PUT (AAS was the exact bug you flagged) now use it so MutationSink.HandleMutation skips the feed insert while still writing the same history row. Also fixed a related root cause - jsonb_agg over MultiLanguageProperty values had no ORDER BY, which could make an identical submodel look changed on read-back.

aaronzi (approved):

PR #616, opened 2026-08-27
Preserve Complete JWT Claim Values During ABAC Evaluation

Martin187187:
## Description of Changes

This pull request aligns the BaSyx ABAC implementation with the current IDTA-01004 Part 4 clarification proposals for nested JWT claims, indeterminate formula evaluation, `ATTRIBUTES`, and CREATE/UPDATE authorization.

The original version of this PR preserved structured JWT claims by serializing arrays and objects as JSON text. That workaround is no longer used. It could not guarantee exact array membership or unambiguous nested selection and would have made authorization dependent on a BaSyx-specific serialization convention.

Instead, verified JWT claims retain their JSON types throughout evaluation. Nested values are selected explicitly with `CLAIMPATH`, string-array membership is exact, invalid runtime values remain indeterminate, and writes are authorized against the state defined by the corresponding operation.

## IDTA specification alignment

This implementation follows four open proposals against the `IDTA-01004-3-2_Working` branch:

| IDTA issue | Specification PR | BaSyx alignment |
| --- | --- | --- |
| [#55: Support for nested/multivalued claims](https://github.com/admin-shell-io/aas-specs-security/issues/55) | [#102: Support nested and multi-valued token claims](https://github.com/admin-shell-io/aas-specs-security/pull/102) | Adds `CLAIMPATH`, strict claim typing and exact string-array membership. |
| [#96: Relationship between `OBJECTS` and `FORMULA`](https://github.com/admin-shell-io/aas-specs-security/issues/96) | [#103: Define indeterminate formula evaluation](https://github.com/admin-shell-io/aas-specs-security/pull/103) | Introduces three-valued evaluation so unresolved fields, missing claims and invalid casts cannot become permissions through negation. |
| [#97: Purpose of `ATTRIBUTES`](https://github.com/admin-shell-io/aas-specs-security/issues/97) | [#104: Clarify the purpose of `ATTRIBUTES`](https://github.com/admin-shell-io/aas-specs-security/pull/104) | Treats declared claims as PIP/token-issuer metadata and resolves them lazily when an evaluated formula branch uses them. |
| [#98: Semantics of `FORMULA` for CREATE](https://github.com/admin-shell-io/aas-specs-security/issues/98) | [#105: Define FORMULA contexts for CREATE and UPDATE](https://github.com/admin-shell-io/aas-specs-security/pull/105) | Evaluates CREATE against staged submitted state and UPDATE against both current and prospective state before commit. |

[...]

PR #654, opened 2026-09-03
Add hierarchical Submodel and SME query support to AAS Environment

Martin187187:
## Summary

AAS Environment shell queries can now evaluate conditions on referenced Submodels and SubmodelElements while respecting each resource's access rules. Caller conditions, response filters, projections, and public list selectors use a common authorization-aware query path across repositories, registries, and discovery services.

The PR also includes the claim-evaluation changes from #616, pins one security policy version per request, corrects object and route coverage, and makes nested casts and regular-expression evaluation safe for restricted data.

### Hierarchical queries and authorization

- Shell queries can reference `$sm` and `$sme` fields in Submodels associated with an AAS. `$match` correlates enclosed conditions to the same referenced Submodel hierarchy.
- Caller expressions evaluate against the data visible under the applicable policy, including fields inside negation, nested casts, date-part operations, related-resource conditions, and response filters. Hidden values cannot be used to select otherwise visible resources.
- An immutable request-scoped authorization session retains the policy version, claims, trusted global attributes, and simplification options. Policy activation during a request does not switch that request to another policy version.
- Per-resource access views preserve each complete allow-rule alternative and its fragment restrictions. Caller projections are combined with security restrictions after authorized resource selection.
- `$aas` IDENTIFIABLE grants cover AAS data and references; they do not grant access to referenced Submodel content. SME `REFERABLE` grants consistently cover the target and its descendants at path-segment boundaries. `FRAGMENT` objects restrict the specified fields.
- Direct and semantic route authorization use consistent mounted paths, including parent-segment normalization and resource/representation boundaries.
- CREATE and UPDATE authorization checks evaluate staged state before commit; UPDATE checks current and prospective state.

The shared path covers AAS and Submodel query/list endpoints, Concept Description lists, AAS and Submodel Registry query/list endpoints, DTR and Basic Discovery lookups, and AAS Environment hierarchical queries. Public selectors use the same visibility checks. Basic Discovery global asset identifier lookups use the corresponding semantic field and join plan.

### Claims and BPN handling

- JWT claims retain their JSON types. `CLAIM` addresses top-level values; `CLAIMPATH` uses a non-empty RFC 6901 JSON Pointer.
- Uncast scalar claim comparisons require strings. Claim casts accept the corresponding JSON type rather than coercing arbitrary strings, arrays, or objects.
- `$contains` with `CLAIMPATH` as its first operand tests exact string-array membership. Other accepted string forms retain substring semantics. The branch-only `$in` form is removed.
- Lazy attribute resolution preserves true, false, and indeterminate results, including through negation and `$match`.
- BPN enrichment copies the claim map before injecting a nonempty `Edc-Bpn` header. The header takes precedence over the signed claim; an absent header leaves the signed claim available. Deployments must continue to prevent untrusted callers from supplying this header directly.

### Safe query evaluation and policy copying

- Fallible field casts validate and convert the same textual input. Nested casts preserve visibility checks, avoid unsupported PostgreSQL source/target casts, and generate SQL whose size grows linearly with nesting depth.
- Invalid cast inputs and invalid regex patterns produce no-match/indeterminate results rather than database cast/regex errors. Regex evaluation retains PostgreSQL's regex operator for query planning.
- External queries and individual policy expressions are limited to **64 JSON container levels and 8,192 JSON tokens**, including keys and delimiters. A query's budget includes its response filters. Oversized query requests return HTTP 400.
- Internally combined policies are structurally deep-copied without reapplying those external input limits. Copying preserves fragment scope, indeterminate markers, formulas by right, and request isolation. Many individually valid allow rules therefore retain their grants after combination.

## Database and upgrade requirements

This PR adds **one production migration: `database/patches/1_1_19.sql`**, registered in the Configuration Service, and advances the schema to **v1.1.19**.

- Requires **PostgreSQL 16+** for `pg_input_is_valid`.
- Installs `basyx_safe_regex_pattern(text)` and `basyx_validated_cast_input(text, text)`.
- Run the Configuration Service to apply the migration before starting updated services. These helpers are required in production; they are not test-only. The patch does not rewrite stored AAS data.

### CatenaXample and DTR

The shipped CatenaXample already emits scalar `role` and `Edc-Bpn` claims. Its updated policies use exact role equality instead of substring matching. Its tested flows need no additional Compose, trust-list, or Keycloak change beyond the database upgrade.

**Existing DTR installations adopting the updated example policy must migrate their role mapper and policy together.** The new policy expects individual role strings in an array:

```json
{"role":["view_digital_twin","add_digital_twin","update_digital_twin","delete_digital_twin"]}
```

Update the live `role` mappers for `aas-env` and `basyx-ui` to `multivalued: true`, store user role attributes as separate values, and obtain fresh tokens when activating the revised policy. Replacing the realm JSON and restarting with `--import-realm` does **not** update an existing realm; use the normal administrative migration process. See [Keycloak startup import behavior](https://www.keycloak.org/server/importExport#importing-a-realm-during-startup).

| DTR policy | Token role format | Provider role grant |
| --- | --- | --- |
| Old policy | Existing comma-separated scalar string | Continues matching canonical roles |
| New policy | Old scalar string | Does not match |
| Old policy | New array | Does not match |
| New policy | Array of individual roles | Exact membership matches |

An array containing one comma-separated string does not satisfy the new membership check. DTR defaults to importing its policy file on startup, so deploying the new policy with an unchanged persistent realm can immediately remove provider grants. Existing old policy/scalar combinations retain their substring semantics.

### Custom-policy compatibility

- Audit parent `REFERABLE` grants: descendants may now receive rights that previously applied inconsistently. Narrow grants if that breadth is unintended.
- Add explicit Submodel/SME grants where a deployment previously relied on `$aas` access for referenced content.
- Align claim types with comparisons and casts; replace branch-only `$in` with `$contains` using `CLAIMPATH` first.
- Expect fewer results where caller conditions previously observed hidden fields, and rejected writes where the prospective state violates authorization.
- Reduce externally supplied expressions exceeding the input limits. There is no new 512-rule policy limit.
- With `policyFileImport=if_missing` or `never`, changing a mounted file may not replace the active persisted policy; import and activate the intended revision explicitly.

Endpoint signatures are unchanged, but authorization outcomes and accepted expression/claim forms have the changes described above.

## IDTA Part 4 scope

This implementation uses the documented **BaSyx policy dialect** and does not claim complete IDTA Part 4 conformance. `CLAIMPATH` is an extension. Existing indeterminate semantics remain: `OR(true, indeterminate)` and `NOT(AND(false, indeterminate))` evaluate to true, differing from Part 4's whole-expression invalidity rule. A false or indeterminate final result does not grant access; a separately true allow rule may grant it. Full conformance requires a separate evaluator and migration decision.

See the [query-language compatibility notes](https://github.com/eclipse-basyx/basyx-go-components/blob/e44fb177921a360abbbb6a018b9ea803372346bd/docu/query_language/README.md#policy-dialect-and-deployment-compatibility), [hierarchical-query authorization guide](https://github.com/eclipse-basyx/basyx-go-components/blob/e44fb177921a360abbbb6a018b9ea803372346bd/docu/query_language/aas_hierarchy_authorization/README.md), and [security documentation](https://github.com/eclipse-basyx/basyx-go-components/blob/e44fb177921a360abbbb6a018b9ea803372346bd/docu/security/README.md).

## Validation

Added or extended regression coverage includes hierarchical joins and `$match` correlation, related-resource grant correlation, hidden-field conditions and nested operands, route/object boundaries, response filters and selectors, typed claims, current/prospective write authorization, invalid casts/regex, and immutable policy copying.

At `e44fb177`:

- Security and grammar package tests passed, including 128-, 512-, and 1,024-rule cases and copy-isolation checks. The 512/1,024-rule regression was reproduced failing before the structural-copy fix.
- A production HTTP/database regression imports, validates, and activates 1,024 distinct READ rules, then confirms that the first and last granted shells are queryable while an unauthorized shell remains hidden.
- Complete AAS Environment security, Submodel Repository integration, Submodel Repository query integration, and DTR security suites passed.
- Repository lint, vet, formatting, and changelog validation passed.

CatenaXample marker and Tractus-X compatibility suites also passed during the preceding rollout review. The Tractus-X harness has ABAC disabled; it supports API compatibility, not security correctness. GitHub checks remain the source of truth for CI on the current head.

RPS benchmarking is outside this review's acceptance scope. The regressions establish bounded-expression and many-rule correctness, not unlimited policy/data scale or unchanged latency. Customer-specific policies, persistent-realm upgrades, and actual EDC/proxy trust configurations still require deployment validation.


internal/aasrepository/api/api_asset_administration_shell_repository_api_service.go
```diff
@@ -134,6 +140,28 @@ func (s *AssetAdministrationShellRepositoryAPIAPIService) QueryAssetAdministrati
 	}), nil
 }
 
+func (s *AssetAdministrationShellRepositoryAPIAPIService) queryContext(ctx context.Context, query grammar.Query) context.Context {
+	queryCtx := auth.MergeQueryFilter(ctx, query)
+	if s.aasHierarchyQueriesEnabled {
+		return grammar.ContextWithAASHierarchyQueries(queryCtx)
```
aaronzi: **High:** This enables caller-controlled `$sm`/`$sme` predicates after merging only the AAS authorization filter. The generated hierarchy `EXISTS` reads Submodel and SME tables directly, without applying the effective Submodel/SME `READ` policy. A principal who may query an AAS but may not read one of its referenced Submodels can use equality, prefix, regex, or numeric predicates as an oracle and infer protected values from whether the AAS is returned. Please intersect the hierarchy scope with the effective Submodel/SME `READ` filter (or reject hierarchy predicates when that cannot be enforced), and add an ABAC integration test with an allowed AAS referencing a denied Submodel.

internal/common/model/grammar/logical_expression_to_sql.go
```diff
@@ -306,7 +346,16 @@ func joinPlanConfigForAAS() JoinPlanConfig {
...
 				},
 			},
-		},
+		}, submodelHierarchyJoinRules("submodel", existsJoinRule{
+			Alias: "submodel",
+			Deps:  []string{"aas_submodel_reference_key"},
+			Apply: func(ds *goqu.SelectDataset) *goqu.SelectDataset {
+				return ds.Join(
+					goqu.T("submodel").As("submodel"),
+					goqu.On(goqu.I("submodel.submodel_identifier").Eq(goqu.I("aas_submodel_reference_key.value"))),
```
aaronzi: **Medium:** The hierarchy join treats every reference key whose value equals a local Submodel ID as the identity key. The reconciliation logic only recognizes `KeyTypesSubmodel`, and permissive verification can persist other key types. As a result, a `GlobalReference` or other non-Submodel key with a colliding value falsely associates the AAS with that Submodel. Please constrain the join to a model reference and `aas_submodel_reference_key.type = KeyTypesSubmodel` as appropriate, with a collision test.

internal/common/model/grammar/query_field_roots.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	for _, match := range expression.Match {
+		if field, found := findMatchExpressionField(match, roots); found {
+			return field, true
+		}
+	}
+	return findExpressionOperandField(
+		[]ComparisonItems{expression.Eq, expression.Ne, expression.Gt, expression.Ge, expression.Lt, expression.Le},
+		[]StringItems{expression.Contains, expression.StartsWith, expression.EndsWith, expression.Regex},
+		expression.BoolCast,
```
aaronzi, line 95: **Medium:** This recognizes a hierarchy field inside a top-level `$boolCast`, so the API accepts it, but `LogicalExpression.EvaluateToExpression` handles standalone `$boolCast` without the collector-aware path. I reproduced `$boolCast: {$field: "$sme.Enabled#value"}` generating an outer predicate that references `property_element` without an `EXISTS` or SME path binding; PostgreSQL will fail with a missing FROM-clause entry and the endpoint returns 500. Please route standalone casts through the collector-aware evaluation and add `$sm`/`$sme` regression coverage.

internal/common/model/grammar/logical_expression_to_sql.go
```diff
@@ -1226,6 +1434,7 @@ func buildInlineExistsExpression(resolved []ResolvedFieldPath, predicate exp.Exp
 		for _, a := range plan.ExpandedAliases {
 			mapped := existsAliases[a]
 			sql = strings.ReplaceAll(sql, "\""+a+"\".", "\""+mapped+"\".")
+			sql = strings.ReplaceAll(sql, a+".", mapped+".")
```
aaronzi: **Medium:** This raw replacement runs over fully rendered SQL, including user string and regex literals. I reproduced a comparison of `$sme.Test#value` with `property_element.literal`; the generated predicate silently compares against `property_element__exists.literal`. Values containing other alias prefixes are affected in the same way. Please rewrite aliases structurally before rendering, or render prepared SQL so literals remain in arguments, and add a regression test containing an alias-like literal.

aaronzi (changes requested):
Thanks for implementing the hierarchical AAS query support and for documenting the intended correlation semantics. I reviewed the query planner, persistence joins, API gating, authorization behavior, and coverage. The overall direction is useful, but I found one cross-resource authorization issue and three correctness problems that should be resolved before merging. I added the details inline.

internal/aasrepository/api/api_asset_administration_shell_repository_api_service.go
Martin187187: Thanks, confirmed. The current implementation merges only the effective AAS READ filter before enabling hierarchy queries. A `$sm` or `$sme` condition is then evaluated through a correlated `EXISTS` that reads referenced Submodel and SME rows without applying their effective READ visibility. Although `/query/shells` returns only AAS objects, its result set becomes an oracle for protected related-resource values.

For example, if Alice may read AAS-A, AAS-A references Submodel-B, and Alice may not read Submodel-B, a condition such as `$sme.Secret#value = "yes"` must not cause AAS-A to be returned only when the hidden value is `yes`. The same applies to prefix, regex, inequality, numeric, cast, negated, and `$match` predicates.

The intended authorization semantics are resource-specific:

- AAS READ alone permits `$aas` conditions only. It never implies `$sm` or `$sme` visibility.
- A `$sm` condition may evaluate only referenced Submodels admitted by the effective Submodel READ rule's ACL gates, object target, and main formula.
- A `$sme` condition may evaluate only referenced SMEs admitted by effective Submodel READ or explicit Referable/Fragment object grants and their main formulas. An exact Referable or Fragment object grant exposes only its target. Descendants are included only when the policy explicitly defines subtree semantics; an `idShortPath` string prefix must not create inheritance implicitly.
- Access-rule `FILTER` and `FILTERLIST` conditions retain their existing response-reconstruction semantics and do not reduce which fields can participate in caller conditions. They are distinct from a `FRAGMENT` entry under `OBJECTS`, which is an authorization target.
- The outer AAS and related-resource grants may originate from separate allow rules. They do not need to be declared in one combined rule.
- Every related-resource guard and caller predicate must apply to the same correlated referenced Submodel and, for `$sme`, the same SME row. Independent `EXISTS` clauses would be insufficient.
- `$aas#submodels[]` is a structural correlation edge, not a third implicit authorization target. If an AAS rule explicitly uses that fragment in its condition or reconstruction filters, that rule still has its normal effect.

Consequently, a caller with access to AAS-A and SME-A inside a referenced Submodel-B, but no general access to Submodel-B, may use SME-A as a condition but may not use `$sm` fields or other SMEs in Submodel-B as conditions. The SQL join through Submodel-B is allowed only to establish ownership and the AAS reference; it must not itself grant or evaluate general Submodel-B data.

PR #659, opened 2026-09-04
QL: Querying the text of a MultiLanguageProperty (#546)

### LINKED ISSUE #546: Query SubmoduleElement of type MultiLanguageProperty
I want to query for a SubmoduleElement where the value is of type MultiLanguageProperty (query the text not the language). 
I tried this query which results in an empty response. 
```json
{
    "$condition": {
        "$and": [
            {
                "$contains": [
                    {
                        "$field": "$sm#idShort"
                    },
                    {
                        "$strVal": "TechnicalData"
                    }
                ]
            },
            {
                "$contains": [
                    {
                        "$field": "$sme.GeneralInformation.ManufacturerProductDesignation#value"
                    },
                    {
                        "$strVal": "e"
                    }
                ]
            }
        ]
    }
}
```
Matching the value together with #language did not lead to any change.
Are queries for MultiLanguageProperty not implemented in Basyx Go or am I missing something?

gintephSICKAG:
# Pull Request Template

## Description of Changes

Implements querying the text of a MultiLanguageProperty.

## Changelog

Select exactly one option:

- [x] I added a Changie fragment under `.changes/unreleased` for every notable user-visible or security-related change.
- [ ] This pull request has no notable user-visible or security effect; a reviewer may apply the `no-changelog` label.

## Related Issue

Closes #546 

## AAS Files Used for Testing

No special AAS file is necessary, every submodel with a MultiLanguageProperty can be used.


aaronzi (approved):
Thanks for the addition to the query language implementation :)

I slightly improved the query language `example.md` file to make it a bit clearer what response to expect when querying MLPs.

laysiSICKAG:
Hi @aaronzi 
I just saw your clarification commit for `docu/query_sprache/examples.md`.
Our initial description wasn't very precise, sorry about that. However, I think your change might still lead to a misunderstanding of the statement.
Your statement is correct when using an `$and` of `#value` and `#language`.
However, when we created that changes, we actually wanted to query also for a specific language, but did that using a `$match` on `#value` and `#language`. Could you please verify that using `$match` operator this actually works and we haven't missed something? Then we might be able to refine the examples and description a bit further.

aaronzi:
Hi @laysiSICKAG, thanks for clarifying. You’re right: I checked the SQL generation, and $match combines both conditions against the same language entry. With the example data, `"Capteur"` combined with `#language = "fr"` matches, while the same text combined with `#language = "en"` does not. $and evaluates those conditions independently.
My clarification only explained the $and behavior and left out the supported $match case. We should show an explicit $match example and add positive and negative integration cases for French and English.
There is one separate specification question: the BNF accepts this expression, but the [QL specification’s “Match of Elements in Lists” section](https://industrialdigitaltwin.io/aas-specifications/IDTA-01002/v3.2/query-language.html) describes $match as requiring an explicit `[]` context. These MultiLanguageProperty fields have no such context. That doesn’t prevent our implementation from working, but we should clarify this upstream before describing this particular usage as standards-compliant.

laysiSICKAG:
Thank you @aaronzi for the confirmation and for pointing out this specification question.
Would you be willing to raise this upstream and share the outcome here once you receive feedback?

aaronzi:
Yes, I will discuss this with @Martin187187 and report back here.

Martin187187:
Hi @aaronzi and @laysiSICKAG, 
this is also the correct behavior from the specification perspective: $match evaluates #value and #language against the same language/text entry of a MultiLanguageProperty. Text containing "Capteur" therefore matches with "fr", but not with "en". $and evaluates these conditions independently.

I’ve opened [#671](https://github.com/eclipse-basyx/basyx-go-components/pull/671) with explicit examples and positive and negative French/English integration cases. All six cases passed, confirming that the implementation behaves as intended.

PR #672, opened 2026-09-11
[Experimental] Adds ReBAC to BaSyx Components

FriedJannik:
## Description of Changes
This pull request introduces experimental resource-bound access control (ReBAC) across multiple services, allowing fine-grained authorization for key resource types. It also updates the initialization and configuration logic for security and database schema management to support this new feature. The changes are primarily focused on enabling and integrating resource-bound access and refactoring security setup for consistency.

**Experimental Resource-Bound Access Control:**

* Experimental resource-bound access control (ReBAC) is now available for concrete AAS, Submodel, SubmodelElement, Concept Description, Registry, and Discovery resources. This includes support for ownership, group principals, inheritance, ABAC fallback, filtered collection reads, capability preflights, and single-use share links. It is enabled via new `EXPERIMENTAL_REBAC_*` environment variables, and requires running the configuration service to apply database schema `v1.1.20`. Previous ReBAC environment variable names are deprecated.

**Security Configuration Refactoring:**

* All services now use the new `SetupConfiguredSecurity` function instead of the previous `SetupSecurityWithABACRepository` for initializing security, streamlining the configuration and supporting the new access control model. [[1]](diffhunk://#diff-51f1bcc3811289797b18916ce371dcc6c81a4339f3e95b19a845700bde74951aL231-R237) [[2]](diffhunk://#diff-1844897dd4209995585aa37d15898659999ecf8a0308b478a7d921bac66d8394L139-R139) [[3]](diffhunk://#diff-2a8654a4d8aaf70e328f028635529ccc7c5398862bc648fc096170e6de1f2d31L180-R188) [[4]](diffhunk://#diff-1d2d3f2793e928e363967576f79c827eca5aa798a58532c110a47b08a5bd35f8L136-R136) [[5]](diffhunk://#diff-07aced2e065509d52ffe396c98823625aee1254dd1bb0077280900f546add837L132-R132) [[6]](diffhunk://#diff-ee38c16347b18545d4fb956226c6b9501a83ca5636f4b25bf4d85ecba90535adL163-R163) [[7]](diffhunk://#diff-00272267017056a3c6fcb8023ebc489db7f3e74ab006a0056f33f110da757179L119-R119) [[8]](diffhunk://#diff-7331a6d6a09a0d4f4bcc214a46a4515d8faf90abed5c6a1359a2af5c7a3998d7L138-R138) [[9]](diffhunk://#diff-76adea4f21342a9c4ef52d3fb7dfe2260bdfe14cb167d86c398cfb121b42d6dbL204-R204)

**Description Service Initialization:**

* The description API services for all relevant components now accept a `ResourceBoundEnabled` flag from the configuration, allowing them to activate or deactivate resource-bound access features as needed. [[1]](diffhunk://#diff-51f1bcc3811289797b18916ce371dcc6c81a4339f3e95b19a845700bde74951aL231-R237) [[2]](diffhunk://#diff-1844897dd4209995585aa37d15898659999ecf8a0308b478a7d921bac66d8394L129-R129) [[3]](diffhunk://#diff-2a8654a4d8aaf70e328f028635529ccc7c5398862bc648fc096170e6de1f2d31L180-R188) [[4]](diffhunk://#diff-07aced2e065509d52ffe396c98823625aee1254dd1bb0077280900f546add837L122-R122) [[5]](diffhunk://#diff-ee38c16347b18545d4fb956226c6b9501a83ca5636f4b25bf4d85ecba90535adL153-R153) [[6]](diffhunk://#diff-00272267017056a3c6fcb8023ebc489db7f3e74ab006a0056f33f110da757179L109-R109) [[7]](diffhunk://#diff-7331a6d6a09a0d4f4bcc214a46a4515d8faf90abed5c6a1359a2af5c7a3998d7L128-R128) [[8]](diffhunk://#diff-76adea4f21342a9c4ef52d3fb7dfe2260bdfe14cb167d86c398cfb121b42d6dbL191-R191)

**Database Schema Management:**

* The configuration service adds support for a new database schema version (`v1.1.20`), ensuring that the database can be upgraded to support resource-bound access control. The previous patch for `v1.1.19` is now explicitly versioned.

These changes lay the groundwork for more granular, resource-specific access control, while maintaining compatibility with existing ABAC mechanisms and ensuring a smooth migration path for deployments.
## Changelog

Select exactly one option:

- [x] I added a Changie fragment under `.changes/unreleased` for every notable user-visible or security-related change.
- [ ] This pull request has no notable user-visible or security effect; a reviewer may apply the `no-changelog` label.


docu/security/ABAC_REBAC_COEXISTENCE.md
FriedJannik: Update this Documentation

internal/common/security/resource_bound_runtime.go
```diff
@@ -0,0 +1,422 @@
...
+					return nil, stateErr
+				}
+				if stateErr = submodelState.checkTarget(r.Context(), tx, matching); stateErr != nil {
+					return nil, stateErr
+				}
+			}
+		}
+	}
+	if right == grammar.RightsEnumUPDATE || right == grammar.RightsEnumDELETE {
+		if err = state.checkDescendants(r.Context(), tx, target); err != nil {
```
aaronzi: For me, this is a blocker: an element-level PUT of a container only checks UPDATE on the existing descendants, but the persistence handlers then delete/reinsert its children. This allows an editor with UPDATE but no DELETE to PUT an empty collection and remove a protected child. An editor with UPDATE but no CREATE can also add a new child this way. Both cases returned 204 in the HTTP reproductions instead of 403.

The delete/reinsert path can also remove the child's access binding, and the context-free insertion does not create a binding for the replacement. Could we reuse the lifecycle reconciliation already used for whole-Submodel PUT here, checking CREATE/UPDATE/DELETE separately in the write transaction and preserving or creating the bindings accordingly? Please cover the equivalent collection, list, entity and annotated-relationship replacement paths with negative tests.

internal/common/security/resource_bound_query.go
```diff
@@ -0,0 +1,393 @@
...
+	if err = rows.Err(); err != nil {
+		return nil, fmt.Errorf("REBAC-STATE-ROWS %w", err)
+	}
+	return snapshot, nil
+}
+
+func (state *boundRequest) queryFilter() *QueryFilter {
+	formula := boolExpression(true)
+	qf := &QueryFilter{Formula: &formula, FormulasByRight: map[grammar.RightsEnum]grammar.LogicalExpression{state.right: boolExpression(true)}, Filters: FragmentFilters{}}
+	qf.Filters["$sme"] = NewFragmentFilterPredicate(boolExpression(true), true)
```
aaronzi: I think we need equivalent mandatory authorization for embedded Submodel Descriptors here. Currently, a reader with READ on an AAS Descriptor but no permission on its nested Submodel Descriptor gets 403 from the direct child GET, while GET of the parent returns 200 and includes that child's identifier, idShort and endpoint URL. I reproduced this with a real, associated AAS/Submodel.

The parent projection can take the unfiltered descriptor path because no $aasdesc#submodelDescriptors[] filter is present. This is a security blocker for me. Could we enforce each embedded descriptor's effective permissions, including its matching Submodel and containing descriptor context, across parent GET/list/query projections? This should not depend on a policy having an explicit FILTER clause; the optimized JSON aggregation path needs the same protection.

internal/common/security/resource_bound_runtime.go
```diff
@@ -0,0 +1,422 @@
...
+	}
+	if state.fallback.Allowed {
+		state.policyID = state.fallback.PolicyID
+		return nil
+	}
+	return boundError(http.StatusForbidden, "ABAC collection action not granted")
+}
+
+func (state *boundRequest) checkDescendants(ctx context.Context, db boundQueryer, target boundTarget) error {
+	if target.Kind != "submodel" && target.Kind != "sme" {
```
aaronzi: This early return leaves descriptor containment outside the descendant checks. I reproduced a case where DELETE of a nested Submodel Descriptor returns 403, but DELETE of its containing AAS Descriptor returns 204 and removes the protected child through the parent deletion.

For me, this is another security blocker: permission on the parent must not bypass the child's separate DELETE permission. Could we extend the transactional lifecycle checks to descriptor descendants, including children removed or replaced through a parent update, and add a regression test where both the direct and parent-mediated deletion are denied?

internal/common/security/resource_bound_runtime.go
```diff
@@ -0,0 +1,422 @@
...
+	state, err := repo.authorizeRequest(r, tx, target, revision)
+	if err != nil {
+		writeBoundError(w, err)
+		return
+	}
+	state.target = target
+	ctx := context.WithValue(common.WithWriterPostgresReads(r.Context()), boundRequestKey, state)
+	ctx = WithQueryFilter(ctx, state.queryFilter())
+	ctx = ContextWithAuthorizationDecision(ctx, AuthorizationDecision{Result: string(DecisionAllow), PolicyID: state.policyID})
+	next.ServeHTTP(w, r.WithContext(ctx))
```
aaronzi: This keeps the authorization transaction, its connection and the scope lock open while the handler runs. The handler then needs another connection from the same writer pool for its own transaction; reads are explicitly routed to that pool above as well.

With MaxOpenConns(1), a normal GET /submodels stalls: the reproduction showed InUse == 1 and WaitCount > 0 until the request was cancelled. Concurrent requests can exhaust larger pools in the same way. I would treat this as a blocker.

Could we either share one transaction through authorization and persistence, or use a versioned authorization snapshot without retaining a separate connection across the handler? We still need to preserve consistency with concurrent policy changes, so simply releasing the lock would need care. A bounded-pool concurrency regression test would be useful here.

internal/discoveryservice/persistence/PostgreSQLDiscoveryDatabase.go
```diff
@@ -202,20 +217,65 @@ func (p *PostgreSQLDiscoveryDatabase) DeleteAllAssetLinks(ctx context.Context, a
...
+			return err
+		}
 		return common.NewInternalServerError("Failed to store specific asset IDs. See console for information.")
 	}
 	return nil
 }
 
+func (p *PostgreSQLDiscoveryDatabase) mutateAssetLinks(ctx context.Context, aasID string, mutate func(*sql.Tx) error) error {
+	return common.ExecuteInTransaction(p.writerDB, "DISC-MUTATE-STARTTX", "DISC-MUTATE-COMMIT", func(tx *sql.Tx) error {
+		exists, err := discoveryAASExistsTx(ctx, tx, aasID)
```
aaronzi: This existence check runs even when ReBAC is disabled, and discoveryAASExistsTx issues a SELECT ... FOR UPDATE. That adds a database round trip and earlier lock acquisition to every CreateAllAssetLinks/AddAllAssetLinks call, although the ReBAC helpers subsequently do nothing for an inactive request.

One of my requirements for this feature is that existing deployments do not pay for it when it is disabled. Could we guard this preparation with ResourceBoundRequestActive(ctx), or move it into a ReBAC-only helper, so the existing mutation path retains its previous query behavior? A disabled-mode query-count regression test would make that guarantee explicit.

internal/submodelrepository/persistence/submodel_write_database.go
```diff
@@ -651,6 +669,17 @@ func (s *SubmodelDatabase) executeSubmodelReconciliationTx(
 	if !plan.hasLiveMutation() {
 		return nil
 	}
+	creates := make([]string, 0, len(plan.Inserts))
```
aaronzi: These slices and loops also run with ReBAC disabled. ResourceBoundReconcileTx returns immediately in that case, but we have already allocated two slices and copied every inserted/updated path. For large Submodel PUTs this adds O(changed elements) work and memory to an existing setup that has not enabled the feature.

Could we guard the list construction and the call with ResourceBoundRequestActive(ctx)? This is smaller than the security findings, but I would still like it fixed before merging to keep the disabled path unaffected.

internal/common/security/resource_bound_query.go
```diff
@@ -0,0 +1,393 @@
...
+		var err error
+		snapshot, err = repo.compileSnapshot(ctx, db, revision)
+		if err != nil {
+			return nil, err
+		}
+		repo.compiled.Store(snapshot)
+	}
+	policyInput := input
+	policyInput.Claims = withBoundGroupClaims(input.Claims, state.actorPrincipals)
+	for _, policy := range snapshot.policies {
```
aaronzi: I am concerned about the scope-wide work here. Every request evaluates every policy in the scope, any scope revision causes the full snapshot to be recompiled, and expression() adds an SQL branch for each policy allowed for the caller, even for a direct request to one resource.

A synthetic SQL-building test produced about 185 KB / 3,000 bind arguments for 1,000 allowed policies and 1.88 MB / 30,000 arguments for 10,000. That means unrelated resources can substantially increase the cost of a single-resource request.

Could we resolve the relevant effective policy IDs first and cache/evaluate those models by binding and revision, while continuing to reuse the ABAC evaluator? Collections will need a bounded or batched approach. I have high confidence in the growth mechanism, but less confidence in the production impact without a realistic load test; I would like a large-scope benchmark before considering this ready for production.

internal/common/security/integration_tests/resource_bound_integration_test.go
```diff
@@ -0,0 +1,950 @@
...
+
+func (p fallbackProvider) ActiveAccessModel() *auth.AccessModel { return p.model }
+
+func newAccessClient(t *testing.T) accessClient { return newAccessClientWithBasePath(t, "") }
+
+func newAccessClientWithBasePath(t *testing.T, basePath string) accessClient {
+	t.Helper()
+	dsn := os.Getenv("BASYX_REBAC_TEST_DSN")
+	if dsn == "" {
+		t.Skip("BASYX_REBAC_TEST_DSN must point to a database migrated to the current schema")
```
aaronzi: Could we wire this suite into CI with an isolated, migrated PostgreSQL database? It currently skips without BASYX_REBAC_TEST_DSN, and the common workflow only runs ./internal/common/security, which does not include this integration_tests package. I could not find a workflow that runs the new package or sets the variable.

For a security feature, I would like these guarantees checked on every change. The existing suite passes when the database is supplied, but the additional negative probes above exposed gaps it does not cover yet. Please also exercise the resource-bound branch through SetupConfiguredSecurity and the registry/discovery/concept-description service compositions; this harness primarily wires the AAS/Submodel controllers directly.

aaronzi (changes requested):
Thanks for implementing this very useful feature! It already looks very promising. I've found a few issues during the review; please have a look at the inline comments below.

The main concerns for me are:

- Existing setups should remain unaffected when ReBAC is disabled, both in terms of security and performance. There is still an extra locking query on Discovery writes and additional allocation/traversal during Submodel PUT, even when the feature is inactive. I'd like those guarded before merging.
- Authorization needs to hold across the complete operation, including nested resources. The focused tests reproduced container PUTs bypassing child CREATE/DELETE permissions, parent descriptor reads exposing a forbidden child, and parent descriptor deletion removing a child whose direct deletion is denied. These are security blockers for me.
- The authorization transaction currently holds a database connection while the handler asks the same pool for another one. A one-connection pool stalls on an ordinary request; concurrent requests can exhaust larger pools too. This is also a blocker.
- The integration should stay clean and reuse the existing ABAC and persistence mechanisms. Reusing the evaluator is only part of that: I'd like the lifecycle reconciliation and composite filtering paths to enforce the decisions consistently, rather than leaving separate operation paths with different checks.
- The new database-backed ReBAC suite needs to run in CI, including the negative cases and the actual service setup/compositions.

One point where I have lower confidence in the practical impact is scalability. Every request currently evaluates all policies in the scope, and the SQL grows with the policies allowed for the caller. A synthetic test produced about 1.88 MB of SQL and 30,000 bind arguments for 10,000 allowed policies. The growth mechanism is clear, but I have not measured production latency or demonstrated an end-to-end resource-exhaustion case. I'd like us to resolve only the relevant effective policies and maybe do a realistic large-scope benchmark on our infrastructure before considering this ready for production.

I also have not established full performance equivalence for disabled deployments. The two extra-work paths above are concrete source-level regressions, but their latency impact has not been benchmarked. I did not confirm a new authorization bypass with ReBAC disabled; that should not be read as a proof that every existing setup is unaffected.

The existing ReBAC integration suite and the full Submodel Repository integration suite passed locally. The focused negative tests described inline exposed the additional issues. I've distinguished the reproduced blockers from the concerns that still need broader validation.

